PR #208, opened 2015-03-25
Fix warnings

blowmage:
This PR addresses several warnings that occur in the library. These warnings occur in projects that use google-api-client. The warnings addressed are:

* assigned but unused variable
* File.exists? is deprecated
* instance variable not initialized
* shadowing outer local variable

You can see these warnings by enabling the `-w` flag when running the specs.

`$ RUBYOPT=-w rake spec`

(There are several more warnings in the specs that are not addressed in this PR, as they will not show up when simply using the library.)

### DIFF
--- a/lib/google/api_client.rb
+++ b/lib/google/api_client.rb
@@ -490,7 +490,7 @@ module Google
       else
         check_cached_certs = lambda do
           valid = false
-          for key, cert in @certificates
+          for _key, cert in @certificates
             begin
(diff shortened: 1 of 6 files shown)


blowmage:
My editor has trimmed whitespace on the files that were changed. Let me know if you would prefer these changes without modifying whitespace.

blowmage:
Thanks! I look forward to the next release. :)

PR #222, opened 2015-04-17
Ridiculously large refactoring...

sqrrrl:
Yes, this is a very large pull request... no rush, but very much want opinions/reviews.

Been working on this the past few weeks to give the client some much needed TLC. The client was at an architectural dead-end and needed a lot of work. 

*Limitations in the current API client*

*  Too much magic - The ruby client relies on runtime code generation to generate scheme classes. As this produces no documentation, it requires users understand the conventions used during generation.
* Inconsistent conventions - Schema classes are generated using ruby naming conventions. Property names in JSON objects are converted to snake case from camel case. However, there are instances where conversion is not applied or where both forms are accepted. Combined with lack of documentation, this has been the source of many issues.
* Clunky interface - The API client offers only a generic ‘execute’ method which takes the API method, parameters, body, etc as parameters. API calls are much more verbose than comparable client libraries.
* Too many options - The client was originally written to support multiple calling styles, have pluggable transports, etc. These were all academic (in the bad way) exercises that produced complexity instead of simplicity. 
* Dependencies on dead 3rd party libraries - The underlying HTTP library used by the library, Faraday, is itself at a dead-end and is no longer being developed. As this both permeates the library and leaks through to users, creating a sustainable path forward requires breaking things.
* Not on latest standards - The client still is designed for old versions of Ruby that are no longer supported and does not leverage language features that would simplify the code. Ruby 2.0+ is standard across the community.

*Principles of the new client*

* No magic (almost) - Ahead-of-time code generation is used to generate services & types for APIs from discovery services. Generated code is fully documented to the extent information is available in the descriptions.
* Simplicity over Flexibility - Sensible defaults are used and few choices are offered to developers. Developers who need a high degree of flexibility can modify the client as needed or write their own.
* Ruby 2.0+ only. 1.8 and 1.9 branches are officially dead. No need to maintain compatibility at this point.


jptalton:
I noticed that some of the API information is missing.  Specifically some of the private Google API's.   

With the discovery method, we could use hidden APIs and the client would still work.

Does this refactored lib support this still?   How should we create new generated api to include in the lib?  What if google doesn't allow us to expose this API (as its private), how do we fallback to dynamic generation internally?

Thoughts / Suggestions / something i missed :)   thank

sqrrrl:
@jptalton Private APIs are still supported in this version. You still need access to the discovery doc as before, the difference is that you'll need it at development time to generate the API classes. That's a one-time operation and the generated classes can kept with the rest of the app's source, put in a separate repo/gem, whatever.

There are some details at the bottom of the [readme](https://github.com/sqrrrl/google-api-ruby-client#generating-apis), but happy to add more if it would help clarify.

For the ones bundled so far, it's limited to preferred(current) versions of public APIs + a few scragglers like the admin directory API that *should* be marked as preferred.

jptalton:
@sqrrrl Thanks.  that indeed worked out.  

Is there any reason not to include some of these other APIs in the repo?

sqrrrl:
If there are missing public APIs, let me know which ones and I'll look at adding them. Private APIs shouldn't be in here though.

snuggs:
Nicely done @sqrrrl. WOW!

sergiopantoja:
:+1: Great PR! Thank you for your work on this.

icco:
cc @erjohnso

sqrrrl:
@tbetbetbe have any chances to wade through this yet?

MIGRATING.md
```diff
@@ -0,0 +1,171 @@
+# Migrating from version `0.8.x` to `0.9`
+
+Many changes and improvements have been made to the `google-api-ruby-client`
+library to bring it to `0.9`. If you are starting a new project or haven't used
+this library before version `0.9`, see the [README][readme] to get started
+as you won't need to migrate anything.
+
+Code written against the `0.8.x` version of this library will not with the `0.9`
```
tbetbetbe: 'will not work with'

sqrrrl:
Squashed a bunch of commits + separating out core from generated. Will restage pull request.

PR #234, opened 2015-06-05
Fixed status code 2xx to return the success? method as true

Student92:
For example a status code of 204 resulted in "success?" returning false


sqrrrl:
Made obsolete by 0.9 release

PR #258, opened 2015-07-20
updated example so that it works and is more relevant to the new api

jasonwbarnett:


jasonwbarnett:
there may be too much with the drive stuff, but it aligns itself with the Drive example used all over the Ruby API client documentation.

PR #262, opened 2015-07-29
Start date and end date are required

rutgerg:


sqrrrl:
Please don't manually edit files under /generated as these will be overwritten during code generation. This requires a fix to the code generator to check for required parameters.

sqrrrl:
Fixed in 0.9.pre3

PR #272, opened 2015-08-18
fix RequestOptions#merge + test for default options

rusikf:
Here
```
 %w(authorization retries timeout_sec).each do |opt|
```
there is no ```header``` option
Also, all this keys we can get with [Struct#members](http://ruby-doc.org/core-2.2.0/Struct.html#method-i-members) method.
Also, add method for overriding default options.


sqrrrl:
Good catch :)

PR #310, opened 2015-11-20
No longer swallowing errors from block passed to execute

### LINKED ISSUE #309: Errors raised in callbacks are swallowed
Given an instance of `Google::Apis::GmailV1::GmailService`:

```
gmail.get_user_message('me', '12345') { |result,err| if err then puts err.message; raise err; end }
```

will print `notFound` but the re-raised error is swallowed, presumably because of the `proc.nil?` check here: https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/http_command.rb#L111

afn:
Fixes #309

### DIFF
--- a/lib/google/apis/core/http_command.rb
+++ b/lib/google/apis/core/http_command.rb
@@ -91,7 +91,6 @@ module Google
         # @raise [Google::Apis::AuthorizationError] Authorization is required
         def execute(client)
           prepare!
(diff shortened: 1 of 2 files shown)


sqrrrl:
Please make sure you sign the CLA. See [CONTRIBUTING](https://github.com/google/google-api-ruby-client/blob/master/CONTRIBUTING.md)

Issue #319, opened 2015-12-14
http command removes initial query parameters


railscard:
When I'm executing service `http` method using url with query values, `HttpCommand` completely replaces those query values with empty hash. 

For example when I'm using this `http` method to download files using the export links from google drive, instead of this:

```
Sending HTTP get https://docs.google.com/feeds/download/documents/export/Export?id=someDocumentId&exportFormat=docx
```
I get this
```
Sending HTTP get https://docs.google.com/feeds/download/documents/export/Export?
```
I think it's not very convenient,  so we should keep initial query parameters. 
If you agree, I can make a PR (https://github.com/railscard/google-api-ruby-client/commit/f51a4f25a954b5b516ad5dc75129d085828ae4b0)

sqrrrl:
PR welcome. Please make sure to read the contributing guidelines and sign the CLA too.

PR #320, opened 2015-12-14
Allow http command to keep initial query parameters

railscard:
Refers to #319


railscard:
Looks like build failed because of bundler issue (https://github.com/bundler/bundler/issues/3558). We need to update bundler before next build.

PR #329, opened 2016-01-02
Remove $LOAD_PATH.unshift(*) and $LOAD_PATH.uniq! from spec/spec_helper.rb because they do nothing

ysksn:


ysksn:
I signed it!

sqrrrl:
I think you also need to update .travis.yml to change the test command to 'bundler rake spec:all'. It was removed a long time ago due to some other issues running through bundler on travis, but it's likely a non issue these days (and we're now forcing a bundler update each time too.)

ysksn:
sqrrrl
thanks for the advice :)

PR #330, opened 2016-01-05
Add Tempfile to if statement of BaseUploadCommand#prepare!

joker1007:
`Tempfile` is not subclass of `IO`. it is subclass of `Delegator`.
Because of it, `Tempfile` instance is regarded as invalid source.
I think that `Tempfile` is useful to upload to bigquery.
API client should be able to use `Tempfile` object.

### DIFF
--- a/lib/google/apis/core/upload.rb
+++ b/lib/google/apis/core/upload.rb
@@ -82,7 +82,7 @@ module Google
         # @raise [Google::Apis::ClientError] if upload source is invalid
         def prepare!
           super
-          if upload_source.is_a?(IO) || upload_source.is_a?(StringIO)
+          if upload_source.is_a?(IO) || upload_source.is_a?(StringIO) || upload_source.is_a?(Tempfile)
             self.upload_io = UploadIO.from_io(upload_source, content_type: upload_content_type)
             @close_io_on_finish = false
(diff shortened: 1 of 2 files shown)


sqrrrl:
Please add some tests. Doesn't need much, just another block like you see in upload_spec.rb line ~ line 123. Something like:

```ruby
  context('with Tempfile input') do
    let(:file) do
      temp_file = Tempfile.new
      temp_file.write("Hello world\n")
      temp_file.rewind
      temp_file
    end
    include_examples 'should upload'

    it 'should not close stream' do
      expect(file.closed?).to be false
    end
  end
```

Also, it might be worth breaking that expression out to its own method since it's starting to get long -- e.g. 

```ruby
def streamable?(obj) ...
```

joker1007:
@sqrrrl 
Thanks for your comment. :smile: 
I added test case and refactored it.

sqrrrl:
Missing  arg to Tempfile.new. Looks like it needs a base name.

joker1007:
Oops... I'm Sorry
I use Ruby-2.3.0. In Ruby-2.3.0, `Tempfile.new` arguments is not necessary.
I fixed and executed test cases in Ruby-2.2.x too.

PR #331, opened 2016-01-06
Remove net webmockhttpresponse

ysksn:
``` ruby
module Net::WebMockHTTPResponse
```
This module used nowhere. Do we still need it?
thanks :)


sqrrrl:
Please keep for now since I'm still planning on supporting both net::http and httpclient for the near future. That particular bit of code relates to monkey patching webmock for a specific test (broken connection during streaming.)

ysksn:
i see

PR #332, opened 2016-01-07
Update copyright to 2016

ysksn:


sqrrrl:
Can't take this merge. The copyright year stays the same since it establishes the start date for when that particular code was first published and copyrighted. Leaving it as 2015 is appropriate. 

Any new files should start using 2016 going forward.

PR #338, opened 2016-01-19
required 'tempfile'

Nakilon:
Otherwise I get:

    gems/google-api-client-0.9/lib/google/apis/core/upload.rb:104:in `streamable?': uninitialized constant Google::Apis::Core::BaseUploadCommand::Tempfile (NameError)

when do

    service.insert_file(metadata, upload_source: "temp.pdf")


Nakilon:
I signed it!

PR #347, opened 2016-01-26
Cleaner code on regexp

sunboshan:
`/[\w_]+/` is equivalent to `/\w+/`. 
 > /\w/ - A word character ([a-zA-Z0-9_])

Ruby-doc [reference](http://ruby-doc.org/core-2.3.0/Regexp.html#class-Regexp-label-Character+Classes).

This will fix the warning when running `ruby -vr google/apis/drive_v3 -e ''`: 
```
/path/to/gems/google-api-client-0.9.1/lib/google/apis/core/api_command.rb:130: warning: character class has duplicated range: /[\w_]+/
```


sqrrrl:
I think you ended up dropping _ from the set.

PR #367, opened 2016-02-24
Fix Hash value assignment in JsonSupport#getter_fn

joker1007:
current version sends `false` value as `nil` if given request parameter is Hash.
because of it, developler cannot change several options that are boolean type and default `true`.

For example `flatten_results` of Bigquery API,
```ruby
Bigquery.client.insert_job(
  Bigquery::PROJECT_ID,
  {
    configuration: {
      query: {
        flatten_results: false,
        query: "SELECT 1",
      }
    }
  }
)
```

I expects

```json
{"configuration": {"query": {"flattenResults": false, "query": "SELECT 1"}}}
```

But actual request parameter is

```json
{"configuration": {"query": {"flattenResults": null, "query": "SELECT 1"}}}
```

### DIFF
--- a/lib/google/apis/core/json_representation.rb
+++ b/lib/google/apis/core/json_representation.rb
@@ -39,7 +39,7 @@ module Google
             ivar_name = "@#{name}".to_sym
             lambda do |_|
               if respond_to?(:[])
-                self[name] || instance_variable_get(ivar_name)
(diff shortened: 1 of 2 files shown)


sqrrrl:
Thanks, good catch. Please add a test case to make sure this doesn't break in the future.

joker1007:
@sqrrrl I added test cases.

PR #368, opened 2016-02-25
replace multi_json dependency with regular stdlib JSON

yob:
JSON has been part of the ruby stdlib since 1.9.0 (see https://github.com/ruby/ruby/commit/af1c4167), and the latest version of this gem requires ruby 2.0 or better.

Would you consider removing multi_json as a dependency, and just using `require 'json'` instead? Rails recently made a similar change (https://github.com/rails/rails/pull/23453), which has prompted other libraries to consider the same issue (like [simplecov](https://github.com/colszowka/simplecov/issues/458) and [stripe](https://github.com/stripe/stripe-ruby/pull/379)).

See also this recent [blog post](http://www.mikeperham.com/2016/02/09/kill-your-dependencies/).

### DIFF
--- a/bin/generate-api
+++ b/bin/generate-api
@@ -6,7 +6,6 @@ require 'thor'
 require 'open-uri'
 require 'google/apis/discovery_v1'
 require 'google/apis/generator'
-require 'multi_json'
 require 'logger'
 
 module Google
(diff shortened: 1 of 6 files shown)


yob:
I've completed the CLA :white_check_mark:

PR #369, opened 2016-02-25
include ruby 2.3.0 in the CI build matrix

yob:
While working on #368, I noticed that travis isn't running the build on MRI 2.3.0 yet. MRI 2.3.0 was released on 2015-12-25, so it'd be good to ensure compatibility. It's fairly compatible with 2.2, so hopefully no changes are required.


sqrrrl:
thanks :)

Issue #370, opened 2016-02-29
Authorize a google developer account without committing a json or other credential info


masonnl:
Hello @sqrrrl and any other maintainers.

I recently took interest in this project and tried to roll it into our app. One drawback is the lack of documentation on authenticating requests to the api. The [Ruby Google Auth library](https://github.com/google/google-auth-library-ruby#example-web) requires a path to a `google_credentials.json` stored in a specific environment variable. We eventually got it working without an explicitly linked credentials file, but it required a lot of code spelunky-ing.

Our solution instead of committing a json file with our Google Application credentials, were environment variables that essentially provide the same information this api client requires to authenticate. This way we can keep secure credentials outside of our repo and instead rely on the server. The three essential environment variables that we set are as follows:

``` sh
GOOGLE_ACCOUNT_TYPE="YOUR ACCOUNT TYPE" # aka: 'service'
GOOGLE_CLIENT_EMAIL="YOUR GOOGLE DEVELOPER EMAIL"
GOOGLE_PRIVATE_KEY="YOUR GOOGLE DEVELOPER API KEY"
```

With the above environment variables we were able to authorize our service app and make queries to the depths of Google. My only concern is that this is more of a workaround for the Google Auth library... Let me know what you think!

**Proposition:**

Provide an additional authorization process as described in the steps above to the README.md.

sqrrrl:
It's not a workaround :) The use of ENV variables like tat is supported by the google auth library. Agree that docs could be improved here though. It's hard to discover that without looking at the code....

neycodes:
Can this be explained in detail? It is still quite unclear how to authorize with ENV variables.

masonnl:
[...]

PR #372, opened 2016-03-01
Add an authorization technique to the docs using env vars

masonnl:
[Google Auth](https://github.com/google/google-auth-library-ruby) supports environment variables as a means of authorizing service apps. Add this technique to the docs for the benefit of others. Issue [#370](https://github.com/google/google-api-ruby-client/issues/370)


buffym:
:+1: even this PR doc saved me SO much time.

PR #400, opened 2016-04-17
Package fewer files in the gem

teeparham:
* Do not distribute test files
* Do not distribute the intermediate api_names_out.yaml file

> Reduces the packaged gem size from 1.8Mb to 1.6Mb

The practice of not distributing test files with a gem was originally
discussed and adopted by the Bundler team here:
https://github.com/bundler/bundler/pull/3207


icco:
Thanks!

PR #409, opened 2016-05-06
fix rubocop issues

zhublik:


zhublik:
I signed it!

sqrrrl:
These are auto-generated files. They'll eventually get overwritten.

Fix needs to be done either in the template, or maybe as part of scripts/generate (e.g. run rubocop -a after generating)

PR #430, opened 2016-06-22
upgrade representable

pfenig:
Just started using this gem so I don't admit to being an expert, but I am somewhat familiar with representable and this seems like what we are shooting for and now tests pass.   The change is from passing options straight into to_json or to_hash to passing the key :user_options.

### DIFF
--- a/google-api-client.gemspec
+++ b/google-api-client.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '~> 2.0'
 
-  spec.add_runtime_dependency 'representable', '~> 2.3.0'
+  spec.add_runtime_dependency 'representable', '~> 3.0'
   spec.add_runtime_dependency 'retriable', '~> 2.0'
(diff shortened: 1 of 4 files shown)


justCxx:
@pfenig @sqrrrl  It is expected any movement on this proposal?

sqrrrl:
It's fine, I'd just like to bundle this with some other larger changes that affect dependencies (dropping Hurley, upgrading httpclient) but I'm blocked waiting for the next httpclient release that contains an important fix. Should hopefully be soon though.

trentclowater:
Does the latest version of httpclient (v2.8.1, released 2 days ago) contain the fix you were waiting for? I'm using reform which requires representable >= 2.4, so would be very happy to see this PR merged.

sqrrrl:
Yep, will take a look at getting a 0.10 release out in the next week or so.

sqrrrl:
Closing, but I applied these changes in the v0.10 branch. Hope to have that out soon.

PR #446, opened 2016-08-04
allow older thor versions at runtime

alepore:
Hello!
I'm using this gem on a old Rails application where i'm stuck with a thor ` ~> 0.14.1` dependency.
This gem requires `~> 0.19` (from gemspec and also a couple of minor development dependencies) but it also work well with thor 0.14 (specs passes).

So, can we lower the thor dependency?
This will not change anything for almost anyone (pointing at latest 0.x version) but will allow some unlucky old apps to use this gem.

Thanks :)


alepore:
I signed it!

seuros:
this gem don't even need thor at runtime. check #447

sqrrrl:
This is obsoleted by #447.

PR #447, opened 2016-08-09
remove thor as runtime dependency

seuros:
thor is used only to generate the api folder.


seuros:
@sqrrrl . Thank you.

Can i work in removing hurley gem and using httpi instead ?

sqrrrl:
I have another branch in progress where I'm removing Hurley and using HTTPClient directly. I know people like to have flexibility in their client and like wrappers like Faraday & HTTPI, but in practice it's hard to support due to leaky abstractions.

seuros:
I think switching to Httpi will from hurley will be not that difficult. Will you accept it if i do it ?

sqrrrl:
Give it a try off the v0.10 branch in development. So long as all the tests continue to pass and the default adapter is set to HTTPClient I'll consider it. No promises though -- I'm still wary of people changing the adapter and uncovering weird edge cases.

PR #486, opened 2016-10-24
Fixes broken filters URL reference

CoryFoy:
Because of the line wrap in the Rubydoc, the URL for filters documentation ends up with an extra %20A in it, which breaks the link from Rubydoc. This undoes the line break so the documentation link is correct. I left it pointing to the V3 docs, since the V4 docs don't seem to have a reference section yet.


CoryFoy:
I signed the CLA

sqrrrl:
Needs to be fixed in the code generator, otherwise it'll just get overwritten each time those files are generated.

PR #519, opened 2017-01-19
Remove unsupported build matrix entry in Travis

saicheems:
Rails 4.2.0 no longer supports Ruby 2.0.0, so this commit excludes it
from Travis builds.


tcoffee-google:
Not familiar with Travis, but LGTM.

PR #528, opened 2017-02-03
Allow abnormal page token cases for paging methods

tcoffee-google:
Addresses googleapis/toolkit#692.

Manual changes are in:
lib/google/apis/core/base_service.rb
spec/google/apis/core/service_spec.rb


tcoffee-google:
Should be submitted concurrently with https://github.com/googleapis/toolkit/pull/986.

saicheems:
Could you explain, roughly, what's going on w/ the changes in `base_service` and `service_spec` so I can understand a bit better?

tcoffee-google:
In the Ruby client, the `fetch_all` method handles internally the extraction of the page token from the response, but the caller handles passing this page token to the next request. I think the purpose of the changes here is made clearer by looking at the corresponding changes in https://github.com/googleapis/toolkit/pull/986.

tcoffee-google:
Any further comments on this?

saicheems:
lgtm

PR #532, opened 2017-02-08
Bump `representable` dependency to 3.0

### LINKED ISSUE #483: Move representable gem version to >=2.4
Representable is now at v3.0 and has a 2.4 version on 2.x. When trying to include this gem in a project, it conflicted with other gems that are using representable >=2.4.

https://github.com/apotonick/representable/blob/master/CHANGES.md

saicheems:
Also fixes a breaking change introduced between `representable` 2.3 and
3.0. Specifically, `:skip_undefined` is now be passed via the
`:user_options` option.

Resolves #483

### DIFF
--- a/lib/google/apis/core/json_representation.rb
+++ b/lib/google/apis/core/json_representation.rb
@@ -55,7 +55,7 @@ module Google
           def if_fn(name)
             ivar_name = "@#{name}".to_sym
             lambda do |opts|
-              if opts[:skip_undefined]
+              if opts[:options][:user_options] and opts[:options][:user_options][:skip_undefined]
                 if respond_to?(:key?)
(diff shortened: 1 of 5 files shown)


lib/google/apis/core/json_representation.rb
```diff
@@ -55,7 +55,7 @@ def getter_fn(name)
           def if_fn(name)
             ivar_name = "@#{name}".to_sym
             lambda do |opts|
-              if opts[:skip_undefined]
+              if opts[:options][:user_options] and opts[:options][:user_options][:skip_undefined]
```
saicheems: Due to a change in how `representable` represents the options map, `:options` is always present, though `:user_options` is not.

PR #533, opened 2017-02-10
Fixed set_thumbnail for youtube (missing videoId error)

detou:
Forcing JSON for request body, the current version was throwing an error: "Required parameter: videoId"


detou:
I signed it!

sqrrrl:
Those files are automatically generated. If there's an issue with the generation, please file a bug to describe the issue or submit a different PR to fix the generator. But as it is, can't merge this as it'll just get clobbered when we regenerate the APIs.

PR #535, opened 2017-02-10
Add a new impersonate user and update email signature sample

thomasdarde:
As previously told, updating from 0.8 to 0.9 has been a very painful process, with very few real world samples.

This samples show authenticate with a service account, how to impersonate someone , and how to update his email signature

### DIFF
--- a/samples/cli/lib/samples/impersonate.rb
+++ b/samples/cli/lib/samples/impersonate.rb
@@ -1,0 +1,61 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


samples/cli/lib/samples/impersonate.rb
```diff
@@ -0,0 +1,61 @@
...
+  #       --impersonated_email='user@example.com'
+  #
+  class Gmail < BaseCli
+    Gmail = Google::Apis::GmailV1
+
+    desc 'impersonate and update email signature', 'Update the email signature of another user'
+
+    method_option :impersonated_email, type: :string, required: true
+
+    def update_email_signature(new_signature_content)
```
saicheems: This seems to use the same API as the existing sample, could you move this function to that class?

samples/cli/lib/samples/impersonate.rb
```diff
@@ -0,0 +1,61 @@
...
+      gmail = Gmail::GmailService.new
+
+      # You can download a client_secret.json form the service account page
```
saicheems: nit: s/form/from

sqrrrl:
Previous comments aside, lgtm.

PR #550, opened 2017-03-06
Update generator to use batchPath instead of hardcoded path

### LINKED ISSUE #523: Update batch endpoint to get URL from discovery
Discovery docs now include the path to the batch endpoint. This should be used instead of hardcoding it to `/batch`.

sqrrrl:
Fixes #523, though APIs will need to be regenerated after this.

Googlers: See b/32514652 for reference.

### DIFF
--- a/lib/google/apis/core/base_service.rb
+++ b/lib/google/apis/core/base_service.rb
@@ -164,10 +164,10 @@ module Google
           batch_command.options = request_options.merge(options)
           apply_command_defaults(batch_command)
           begin
-            Thread.current[:google_api_batch] = batch_command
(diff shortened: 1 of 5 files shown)


sqrrrl:
Technically this is a breaking change. Previously it was possible to mix calls from different services in a batch, even if it was discouraged by the API design. Using the API specific endpoints will cause that to break. In preparation, added some stricter checks to prevent it at the client level.

If people need it, it's possible to add a check and direct mixed batches to the global batch endpoint, but for now I'm keeping it simple.

ethanbao:
Is this ready for review?

PR #579, opened 2017-05-12
Update README.md

saicheems:


saicheems:
+@omaray

tcoffee-google:
I defer to @omaray, but current wording implies to me that (a) the new client library generators are inherently specific to Cloud APIs, and (b) the old Ruby libraries are unidiomatic (which I'd argue is selling them short).

omaray:
Node.js uses similar language. It worked out. Let's update as-is and always revisit if need be.

saicheems:
Edited the language to match what's on `google-cloud-nodejs`. Feel free to ping me or make a PR to change the wording. I'll roll with this for now.

PR #583, opened 2017-05-19
Allow using v2 error messages + fix to_json signature 

sqrrrl:
Adds opt-in to use v2 error messages (google/google-api-ruby-client#544). Also fixes the `to_json` signature for compatibility with the `json` gem  (google/google-api-ruby-client#576)


saicheems:
lgtm

qingling128:
Hi, Is this ready for merge? :)

PR #586, opened 2017-06-01
Add Hash support to fetch_all

saicheems:
This commit aligns the behavior of `fetch_all` over map responses with
that of arrays (returning a single collection with all entries).

Fixes https://github.com/google/google-api-ruby-client/issues/584

### DIFF
--- a/lib/google/apis/core/base_service.rb
+++ b/lib/google/apis/core/base_service.rb
@@ -69,6 +69,13 @@ module Google
                 break if @max && item_count > @max
                 yield item
               end
+            elsif items.kind_of?(Hash)
+              items.each do |key, val|
+                item_count = item_count + 1
+                break if @max && item_count > @max
+                item = { key => val }
+                yield item
+              end
             elsif items
               # yield singular non-nil items (for genomics API)
               yield items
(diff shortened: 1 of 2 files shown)


saicheems:
Will now behave like this:

```
items = service.fetch_all do |token|
  service.aggregated_autoscaler_list(project, page_token: token)
end

items.each do |name, autoscalers_scoped_list|
  # TODO: Change code below to process each (name, autoscalers_scoped_list) pair:
  puts String(name) + " => " + autoscalers_scoped_list.to_json
end
```

instead of this (before):

```
items = service.fetch_all do |token|
  service.aggregated_autoscaler_list(project, page_token: token)
end

for item in items
  item.each do |name, autoscalers_scoped_list|
    # TODO: Change code below to process each (name, autoscalers_scoped_list) pair:
    puts String(name << " => " << autoscalers_scoped_list.to_json
  end
end
```

lib/google/apis/core/base_service.rb
```diff
@@ -69,6 +69,12 @@ def each
                 break if @max && item_count > @max
                 yield item
               end
+            elsif items.kind_of?(Hash)
+              items.each do |key, val|
+                item_count = item_count + 1
+                break if @max && item_count > @max
+                yield key, val
```
saicheems: Note that this is collected as an array of 2 item arrays

ex: `[ [ "key1", "value1" ], [ "key2", "value2" ] ]`

I'm not sure what the most idiomatic behavior is, please let me know.

saicheems:
Sure, I'll have the next release be a major version bump

PR #593, opened 2017-06-23
Recover from non-Ranged responses without rewinding

janko-m:
Not all IO objects know how to #rewind themselves. For example, Ruby pipes (returned by `IO.pipe`) do implement #rewind, but they will throw an error if you try to call it.

```rb
rd, wr = IO.pipe
wr.rewind # Errno::ESPIPE: Illegal seek
```

But we don't need to rewind and overwrite the IO object if we didn't get the Ranged response we expected, we could instead wait out the content that has already been downloaded, and start appending again once we reached where we left off. This is what this PR does.

This allows us to use the write end of a Ruby pipe as the download destination, which enables me to implement download on-demand (to partially download only the amount bytes needed for an operation, e.g. extracting metadata that are only found at the beginning of the file).

### DIFF
--- a/lib/google/apis/core/download.rb
+++ b/lib/google/apis/core/download.rb
@@ -67,12 +67,11 @@ module Google
         def execute_once(client, &block)
           request_header = header.dup
           apply_request_options(request_header)
+          download_offset = nil
 
-          check_if_rewind_needed = false
           if @offset > 0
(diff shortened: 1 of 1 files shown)


janko-m:
The reason why I switched from `String#length` to `String#bytesize` is explained in https://github.com/google/google-api-ruby-client/pull/592#issue-238070672.

PR #597, opened 2017-07-07
maintenance mode notice

danoscarmike:
Add notice to top of README to highlight that this library is in maintenance mode only (i.e. not under active development).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Google API Client
 
+## Library maintenance
+
+This client library is supported but in maintenance mode only.  We are fixing necessary bugs and adding essential features to ensure this library continues to meet your needs for accessing Google APIs.  Non-critical issues will be closed.  Any issue may be reopened if it is causing ongoing problems.
+
 ## Alpha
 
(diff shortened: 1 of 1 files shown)


danoscarmike:
@remi @geigerj could you please review?

benoittgt:
I'm sorry but why it's in maintenance mode? 🤔

saicheems:
The library has been in maintenance mode for some time - we will continue to support it as we have (client library updates, bug fixes, etc.). This is just spelling out expectations, since we'd like new users to build with the more idiomatic and actively developed [google-cloud-ruby](https://github.com/GoogleCloudPlatform/google-cloud-ruby) client.

TrevorHinesley:
@saicheems I've looked and I don't see how https://github.com/GoogleCloudPlatform/google-cloud-ruby integrates with GMail like this client. Does it?

saicheems:
google-cloud-ruby will not integrate with GMail because it is not a Google Cloud API, you should use this library for that use case.

PR #601, opened 2017-07-17
Make the generator deterministic

saicheems:


lib/google/apis/generator/templates/_method.tmpl
```diff
@@ -15,7 +15,7 @@
 #   <%=  block_comment(param.description, 0, 3) %>
 <%   end -%>
 <% end -%>
-<% for param in api.parameters.values.reject {|p| p.name == 'key'} -%>
+<% for param in api.parameters.values.sort_by {|p| p.name}.reject {|p| p.name == 'key'} -%>
```
saicheems: I wasn't able to figure out how to get this properly sorted in `model.rb`. I tried adding the method:

```
def parameters
  Hash[@parameters.sort]
end
```

under `RestDescription` in `model.rb`, but for some reason sorting the hash caused many other service level parameters to be included... For example, right now only "fields" and "quota_user" are generated, but if I sort the hash in `model.rb`, then other fields like "oauth_token" and "pp" are also included for each method. I couldn't debug that at all, so I explicitly included the sort in the template.

Please let me know if I was missing something obvious.

sqrrrl: There's a line in annotator.rb that does the blacklisting:

            @rest_description.parameters.reject! { |k, _v| PARAMETER_BLACKLIST.include?(k) }

Returning a new hash each time renders that line ineffective. You can go ahead and move that elsewhere though so you can do the sorting in the model instead of the templates.

saicheems: Got it, I moved that line to `model.rb`. I think that's solved the issue.

sqrrrl:
lgtm

PR #605, opened 2017-07-27
Allow automatic method name clash resolution

saicheems:
If the method name is generated from the request type, there is a
possibility of a clash if two methods take the same request type. For
example, if two unique methods take the request type
"CreateRoleRequest", both methods will be generated with the method name
"create_role", which causes generation to fail with an exception.

This commit resolves that issue by first generating a method name from
the request type of each method, and keeping track of which names are
duplicates. For methods that will generate a duplicate name from the
request type, the name is generated from the method ID instead (the
resulting name is guaranteed to be unique).

### DIFF
--- a/lib/google/apis/generator.rb
+++ b/lib/google/apis/generator.rb
@@ -27,8 +27,9 @@ module Google
       Discovery = Google::Apis::DiscoveryV1
 
       # Load templates
-      def initialize(api_names: nil)
-        @names = Google::Apis::Generator::Names.new(api_names || File.join(Google::Apis::ROOT, 'api_names.yaml'))
+      def initialize(api_names: nil, api_names_out: nil)
+        @names = Google::Apis::Generator::Names.new(api_names_out || File.join(Google::Apis::ROOT, 'api_names_out.yaml'),
+                                                    api_names || File.join(Google::Apis::ROOT, 'api_names.yaml'))
         @module_template = Template.load('module.rb')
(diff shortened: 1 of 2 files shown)


sqrrrl:
Mostly looks good, but please check two cases to be sure:

- Please make sure it doesn't inadvertently rename any existing methods when regenerating. 
- Check the case where a conflict is created at a later point in time to make sure the first method remains stable. This has been the source of most conflicts -- things work OK for a while then a method is added in a subsequent release that causes a conflict.

IIUC the code correctly, I think the 2nd case might break unless we lock in the name in the overrides.

saicheems:
I think those cases will be tricky to solve through logic in the annotator by itself...

The only solution to both that I can think of immediately is to, as part of the regeneration step, copy the generated `names_out.yaml` file and use that as input to the next generation (merged with `api_names.yaml`, such that names in `api_names.yaml` take precedence). That way, names are locked forever.

Does that solution seem reasonable to you?

saicheems:
I implemented the solution I suggested above - the generator first loads `api_names_out.yaml`, and then merges `api_names.yaml`. That should eliminate the possibility of a name ever changing once its been generated.

Does that seem reasonable to you? Or is there a better way

saicheems:
Friendly ping, any thoughts on the above approach?

saicheems:
Friendly ping, any thoughts on the above approach?

sqrrrl:
lgtm

PR #609, opened 2017-08-11
Handle recursive hashes

### LINKED ISSUE #608: How to use walletobject for save to android pay api?
I am following https://developers.google.com/save-to-android-pay/samples/quickstart-ruby guide which uses https://github.com/android-pay/s2ap-quickstart-ruby this repository has not been updated to use the new version of google-api-client. I tested it and It still works with the old version  client (<0.8).

I am trying to migrate this to the new version before implementing it in my system. However, I am not sure if the file https://github.com/android-pay/s2ap-quickstart-ruby/blob/master/wobs-discovery.json would work with the new file generation discovery. Currently when I run it I get a 'stack level too deep' error.

Any idea? I'm really not sure where to ask about this.

saicheems:
This fixes the generator for cases where one of the entries of the
"schemas" object returned by the Discovery service is a recursively
defined map.

Fixes #608

### DIFF
--- a/lib/google/apis/generator/model.rb
+++ b/lib/google/apis/generator/model.rb
@@ -55,8 +55,14 @@ module Google
             return 'Fixnum' if format == 'uint64'
             return TYPE_MAP[type]
           when 'array'
+            if items == self
+              return sprintf('Array<%s>', qualified_name)
+            end
             return sprintf('Array<%s>', items.generated_type)
           when 'hash'
+            if additional_properties == self
+              return sprintf('Hash<String,%s>', qualified_name)
+            end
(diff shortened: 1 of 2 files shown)


saicheems:
For reference, see: https://github.com/android-pay/s2ap-quickstart-ruby/blob/master/wobs-discovery.json#L2871

^ `TypedValue` is a map and an object which points to itself.

sqrrrl:
Looks good, but would like to see a test case added for it.

saicheems:
Sure, I'll get to it next week.

hshar7:
@saicheems ping

saicheems:
Fixed in #621

PR #611, opened 2017-08-22
Allow auth retry on Signet::AuthorizationError

saicheems:
Fixes https://github.com/google/google-api-ruby-client/issues/554

### DIFF
--- a/lib/google/apis/core/http_command.rb
+++ b/lib/google/apis/core/http_command.rb
@@ -99,7 +99,7 @@ module Google
               # NotFound, etc
               auth_tries = (try == 1 && authorization_refreshable? ? 2 : 1)
               Retriable.retriable tries: auth_tries,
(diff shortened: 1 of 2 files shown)


saicheems:
I need some advice on how to test this addition... I wasn't able to figure out how to trigger a `Signet::AuthorizationError` in the relevant test code.

Do you have any advice?

sqrrrl:
You can stub out the network call for the token endpoint to test it. Returning a 500 error should trigger it.  You'll have to configure the request with some garbage credentials that are expired for the refresh code to be triggered.

Note that there are two cases where a token can be refreshed and that error thrown. The first is an expired token that is automatically refreshed (`apply!` will auto refresh if the token is expired or near expiring.) There's a second case -- the underlying service returns a 401 error. That will trigger a refresh as well.

Since those two cases are handled differently, should test both.

sqrrrl:
lgtm

PR #613, opened 2017-08-24
Generate all preferred APIs by default

saicheems:
This commit removes support for the `id` flag from the `generate-api`
script. Additional Discovery documents are still supported via the `url`
flag.

The new behavior is to generate clients for all APIs marked as preferred
by the Discovery index if the `from_discovery` flag is `true`.


saicheems:
Will roll with this for now, let me know if you want any changes.

sqrrrl:
There are some actively developed APIs that are not marked as preferred. Specifically, need to make sure all the admin:* APIs are generated as well as drive:v2.

PR #621, opened 2017-10-11
Handle recursive hashes

### LINKED ISSUE #608: How to use walletobject for save to android pay api?

bantini:
Extension of pull request #609 with test case
Fixes #608


bantini:
Signed cla with github email.

bantini:
Signed cla.

bantini:
I signed it!

spec/fixtures/files/test_api.json
```diff
@@ -131,6 +134,49 @@
...
+      },
+      "string": {
+       "type": "string",
+       "description": "The string value."
+      }
+    },
+   "additionalProperties": {
+    "$ref": "RecursiveHat"
+   }
+  },
```
saicheems: Something is weird about the indenting here.

spec/fixtures/files/test_api.json
```diff
@@ -131,6 +134,49 @@
         ]
       }
     },
+    "RecursiveHat": {
```
saicheems: Do tests fail for this schema without the change to `model.rb` and `_class.tmpl`?

bantini: Tests fail without the change to model.rb, but not for _class.tmpl.

spec/fixtures/files/test_api.json
bantini: Fixed in the latest commit.

PR #624, opened 2017-10-18
Update version constraint 'googleauth'

koenrh:
This updates the version constraint of 'googleauth' so that we can use the latest version of that gem. This allows us to use the next major version (i.e. version 2.0) of the 'jwt' gem for which the version constraint was loosened in https://github.com/google/google-auth-library-ruby/pull/119.

Is there any particular reason to pin on a specific minor version, and not on a major version instead?


koenrh:
Superseded by https://github.com/google/google-api-ruby-client/pull/628.

Issue #636, opened 2017-11-28
Is it possible to disable Gzip decompression?


quartzmo:
To resolve GoogleCloudPlatform/google-cloud-ruby#1835, we need to disable the Gzip decompression that is automatically performed by google-api-ruby-client. I believe it is enabled here in [base_service.rb#L403](https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/base_service.rb#L403).

Is it currently possible to do this, or else can the library be updated to allow for this configuration to be modified?

/cc @blowmage @geigerj

geigerj:
@bantini Could you please check into this and provide a work estimate if the library needs to be updated? It's possible to work around this downstream in the `google-cloud-ruby` client but a solution here would be better.

bantini:
@quartzmo Can this issue be closed?

quartzmo:
Yes, thank you!

quartzmo:
@bantini When can we expect a release containing the change in #637?

bantini:
@quartzmo Unfortunately, I am not aware of the release process. 
CC @geigerj @saicheems

saicheems:
Released in 0.17.4

quartzmo:
@saicheems Thank you!

PR #637, opened 2017-11-29
feat: Allow disabling of the gzip decompression using client_options

bantini:
Addresses #636 
Added option to disable gzip decompression through client_options


bantini:
@geigerj Do you think this is ready to merge? Or is there anyone else who should review it?

geigerj:
I think this is ready to merge. @quartzmo to confirm that this new knob meets the downstream requirements.

quartzmo:
This looks great. We are [already setting some `ClientOptions` values](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/service.rb#L48-L50), so it looks very straightforward. Of course, I won't know that it actually solves our problem (retrieving un-gzipped content) until I try it, but I'll do that as soon as I can.

quartzmo:
Sorry, I meant: retrieving the gzipped content and having it stay gzipped.

quartzmo:
Thank you, @bantini!

quartzmo:
I implemented a solution for GoogleCloudPlatform/google-cloud-ruby#1835 based on this change by loading `master` branch of this library via git, and it solves our issue! When can we expect a release containing this change?

PR #642, opened 2017-12-21
Improve language in Logging section of README

bbuchalter:
For readers following the README's sequence of instructions to get logging to STDOUT/STDERR, they will generally add code in the order which it appears in the README. That results in this:
```
Google::Apis.logger.level = Logger::DEBUG
Google::Apis.logger = Logger.new(STDERR)
```

However, this configuration will not have the desired effect in Rails apps, since the newly created logger will not respect the log level which was actually set on the Rails logger. By changing the order in which these code snippets appear in the README, we can help readers avoid this gotcha.

Additionally, I offer a small semantic improvement to help readers better understand the two options they have for changing the default logger.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,23 +281,20 @@ GOOGLE_PRIVATE_KEY="YOUR GOOGLE DEVELOPER API KEY"
 
 The client includes a `Logger` instance that can be used to capture debugging information.
 
-To set the logging level for the client:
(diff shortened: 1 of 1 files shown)


bbuchalter:
Signed

grant:
LGTM.

PR #685, opened 2018-06-04
Note change to normalize_unicode

### LINKED ISSUE #684: ::Google::Apis::ClientOptions.default.normalize_unicode no longer exists
`::Google::Apis::ClientOptions.default.normalize_unicode = true` no longer works as it's now a request option, please, add a note to `MIGRATING.md`

mattwhisenhunt:
Fixes #684.

@Temikus

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -4,11 +4,7 @@
 
 The client no longer normalizes unicode strings in path parameters. This may affect
 some applications using multibyte strings that were previously normalized.:
-To restore the previous behavior, set the following option prior to creating a service.
-
-```ruby
-ClientOptions.default.normalize_unicode = true
-```
+To restore the previous behavior, use the `normalize_unicode` request option.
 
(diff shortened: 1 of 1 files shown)


MIGRATING.md
```diff
@@ -4,11 +4,7 @@
 
 The client no longer normalizes unicode strings in path parameters. This may affect
 some applications using multibyte strings that were previously normalized.:
-To restore the previous behavior, set the following option prior to creating a service.
-
-```ruby
-ClientOptions.default.normalize_unicode = true
-```
+To restore the previous behavior, use the `normalize_unicode` request option.
```
Temikus: Thanks!

Nit:To keep the record in MIGRATING.md straight - is it possible to add a note with a version when it changed to a request option, instead of rewriting history?

Temikus: Ah, fair enough. Thanks for looking into it! :)

PR #714, opened 2018-09-06
Fix people class name

walkersumida:


grant:
This is a generated class, so the change will be overwritten when generated again.
Thus, I can't accept this PR.
https://github.com/googleapis/google-api-ruby-client#generating-apis

walkersumida:
I understand. Thank you for your comment.

PR #730, opened 2018-10-24
Update googleauth dependency

blowmage:
Allow the dependency to use the 0.7.x releases.


quartzmo:
How is this change being tested?

blowmage:
I've tested this against google-cloud-ruby with googleauth 0.7.1 and everything works as expected.

PR #735, opened 2018-11-07
add kokoro configs

TheRoyalTnetennba:


TheRoyalTnetennba:
Windows failures are addressed in https://github.com/googleapis/google-api-ruby-client/pull/736

PR #776, opened 2019-03-19
Deserialize type: Date values

RedBulli:
According to https://www.rubydoc.info/github/google/google-api-ruby-client/Google/Apis/CalendarV3/EventDateTime#date-instance_method the ```date``` should be parsed to a ```Date```, but currently it is only a "yyyy-mm-dd" string.


quartzmo:
As @blowmage feared, this change led to unit test failures in google-cloud-storage (fixed in googleapis/google-cloud-ruby#3417).

blowmage:
The joys of using a library that has never reached 1.0...

PR #778, opened 2019-04-10
Remove third_party directory from gemspec

blowmage:
This directory was used when the gem was patching other gems,
like Hurley. But now this directory has been removed.


TheRoyalTnetennba:
There's still a test for it https://github.com/googleapis/google-api-ruby-client/blob/711dfb83b33c03535076917726956584d5c8bf9a/spec/spec_helper/load_path_spec.rb#L31

PR #789, opened 2019-05-10
Replace mime-types dependency with mini_mime

bborn:
Replaced mime-types with mini_mime. Specs are passing.


bborn:
Signed the CLA

google-api-client.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'representable', '~> 3.0'
   spec.add_runtime_dependency 'retriable', '>= 2.0', '< 4.0'
   spec.add_runtime_dependency 'addressable', '~> 2.5', '>= 2.5.1'
-  spec.add_runtime_dependency 'mime-types', '~> 3.0'
+  spec.add_dependency 'mini_mime', '>= 0.1.1'
```
blowmage: We should match how the other dependencies are declared. Also, why not pin to 1.0?

```suggestion
  spec.add_runtime_dependency 'mini_mime', '~> 1.0'
```

PR #848, opened 2019-11-12
Document Ruby compatibility change in changelog

ivy:
Hello! 👋 This is a quick addition to the changelog to document the recent change in Ruby support.

From what I've been able to verify, it looks like 0.33.0 was the last time Ruby < 2.4 was supported. Introduced in #834 and verified with:

```
$ git tag --contains bc573bf98 | head -n1
0.33.0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,7 +137,6 @@
 
 # 0.33.0
 * Breaking changes:
-  * Dropped support for Ruby versions older than 2.4.
   * Deleted `urlshortener_v1`
 * Backwards compatible changes:
   * Updated `accessapproval_v1beta1`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -137,6 +137,7 @@
 
 # 0.33.0
 * Breaking changes:
+  * Dropped support for Ruby versions older than 2.4.
```
blowmage: I don't know that I would list this as a breaking change. We haven't considered this to be breaking for other gems.

ivy: How about listing it under a new *Backwards incompatible changes* section? While it isn't necessarily a breaking change, it's still a pretty big change in support.

blowmage: We have used "Other" to describe this change in the Google Cloud Platform gems.

https://github.com/googleapis/google-cloud-ruby/releases/tag/stackdriver/v0.16.0

This isn't an API change like most of the changes listed.

ivy: That works for me. 😄

blowmage:
@dazuma You okay with these changes?

PR #867, opened 2020-03-25
decode_response_body empty body handling

androidseb:
A more robust REST API response handling for empty body returns.
The rest API is not supposed to return an empty string since it's invalid JSON, but it can happen in some cases.
This change makes the decode_response_body method more robust to handle this case.

### DIFF
--- a/lib/google/apis/core/api_command.rb
+++ b/lib/google/apis/core/api_command.rb
@@ -87,6 +87,9 @@ module Google
           return super if content_type.nil?
           return nil unless content_type.start_with?(JSON_CONTENT_TYPE)
           instance = response_class.new
+          if body.empty?
+            body = '{}'
+          end
           response_representation.new(instance).from_json(body, unwrap: response_class)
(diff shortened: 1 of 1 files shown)


androidseb:
The specific case where this can happen is when performing a DELETE request on the Google Play Publishing REST API with request like:
DELETE https://www.googleapis.com/androidpublisher/v3/applications/com.my.app.id/edits/my_edit_id_as_number/listings/en-CA/phoneScreenshots?
In this case, the response body is empty ('') instead of an empty JSON object ('{}'), and the decode_response_body function fails.

androidseb:
There is an open issue in the Fastlane project about this too, adding the link for reference since the reported issue show-cases the ruby stack trace of the error:
https://github.com/fastlane/fastlane/issues/16175

androidseb:
The bug is no longer reproducible, it seems the bug in the Google Play REST API was fixed.
Closing this PR now.

PR #2132, opened 2020-12-23
Fixing README.md links

od-c0d3r:
After moving google-api-client into a subdirectory.


od-c0d3r:
@googlebot I signed it!

dazuma (approved):
Thanks!

PR #2342, opened 2021-01-09
Add cause's class to Google::Apis::Error message

mohamedhafez:
Knowing what the original error that was thrown was can be useful, and it would be nice to be able to know at a glance when one of these errors is raised.

### DIFF
--- a/google-apis-core/lib/google/apis/errors.rb
+++ b/google-apis-core/lib/google/apis/errors.rb
@@ -24,7 +24,7 @@ module Google
         @cause = nil
 
         if err.respond_to?(:backtrace)
-          super(err.message)
+          super("caused by #{err.class}: #{err.message}")
           @cause = err
         else
           super(err.to_s)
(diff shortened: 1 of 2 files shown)


mohamedhafez:
@googlebot I signed it!

bajajneha27:
Hi @mohamedhafez , thank you for opening this pull request. I'm trying to understand as to why this change is required. Is there anything specific you're trying to achieve?

mohamedhafez:
It would just be nice/useful to know what caused the error sometimes. You guys already include the `message` of the underlying exception for this reason, but but it would be nice to know the actual class of that underlying exception at a glance as well

bajajneha27:
If you have access to the error object, you can always get the class of it by using `class` method on it. Is there anything I'm missing here?

mohamedhafez:
Sure, in a rescue block I could always do `$!.cause.class` to find out the class of the underlying exception. I just think it's a nice convenience thing to know what the underlying exception class is in the `message` of the main exception object, so it's immediately obvious what went wrong. I mean that's why we are putting in the the cause's `message` in there right? Personally I think sometimes that's not enough to know exactly what went wrong, and this PR adds a little more context. But I suppose it comes down to taste, you are correct that the information is still available even without this PR

bajajneha27:
Hi @mohamedhafez ,
Sorry for the delay in response. While I understand what you mean and that makes sense but I also think that making this a general message won't be a good decision.

NivedhaSenthil:
Closing this PR as the cause is available in the error.

PR #3414, opened 2021-03-23
feat: add opentelemetry instrumentation

robertlaurin:
Provides instrumentation for https://github.com/open-telemetry/opentelemetry-ruby

Original draft PR feedback here https://github.com/googleapis/google-api-ruby-client/pull/3185


google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -355,7 +358,46 @@ def safe_response_representation http_res
...
+          return if @opentelemetry_span
+
+          attributes = {
+            'http.host' => url.host.to_s,
+            'http.method' => method.to_s,
+            'http.target' => url.path.to_s,
+            'peer.service' => url.host.to_s
+          }
+
+          @opentelemetry_span = Google::Apis::Core::OpenTelemetry.instance.tracer.start_span(url.host.to_s, attributes: attributes)
```
robertlaurin, line 372: > In general, what's the best practice for naming spans? This just provides the service name. Should we include the method being called?
https://github.com/googleapis/google-api-ruby-client/pull/3185#discussion_r591982855

That's a good question as this isn't always exactly an HTTP span, if the consumers of this instrumentation also include the [http client instrumentation](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/instrumentation/http_client) these spans will be wrapping the HTTP calls.

For our internal instrumentation we've been using `url.host.to_s` as the span name.

robertlaurin:
@googlebot I signed it!

google-apis-core/lib/google/apis/core/opentelemetry.rb
```diff
@@ -0,0 +1,26 @@
...
+module Google
+  module Apis
+    module Core
+      class OpenTelemetry < OpenTelemetry::Instrumentation::Base
+        install { true }
+        present { true }
+      end
+    end
+  end
+end
```
robertlaurin, line 26: With the [addition of the dynamic upgradeable tracer](https://github.com/open-telemetry/opentelemetry-ruby/pull/708) do we still want to make use of this?

robertlaurin (commented):
hey @dazuma, there was a bit of a delay getting approval for the CLA.

google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -99,20 +101,23 @@ def execute(client)
...
-              # NotFound, etc
-              auth_tries = (try == 1 && authorization_refreshable? ? 2 : 1)
-              Retriable.retriable tries: auth_tries,
-                                  on: [Google::Apis::AuthorizationError, Signet::AuthorizationError, Signet::RemoteServerError, Signet::UnexpectedStatusError],
-                                  on_retry: proc { |*| refresh_authorization } do
-                execute_once(client).tap do |result|
-                  if block_given?
-                    yield result, nil
+            url_host = url.host.to_s
+            OTEL_TRACER.in_span(url_host) do
```
robertlaurin, line 105: We have instrumentation for the underlying http lib (http client), so I think it makes sense to not duplicate all the HTTP attributes on this span. 

This span also serves the purpose of grouping retired network calls under a common parent span.

genebean:
What's the status of this work? Asking because I am very much interested in putting this to use. Also, is there additional code somewhere that would set attributes like is done for the aws_sdk in https://github.com/open-telemetry/opentelemetry-ruby/blob/f0593e7d97c62b10862c382d14f533dbc49be580/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb#L23-L28 ?

bajajneha27:
Hi @robertlaurin , I'm so sorry we completely lost track of this. Is it still updated? Would you like to take a look and update the PR accordingly?

bajajneha27:
Hi @robertlaurin 
I'm closing this PR due to inactivity for now but please feel free to reopen it incase you still want to work on this.

PR #6494, opened 2021-09-14
Properly log errors as errors

ylecuyer:
Atm almost all logs from the core api are debug logs. The default level for the logger is set to warn meaning, we see almost nothing at the beginning and it gets to verbose as soon as we reach the debug level.

This is not practical to debug real errors that why I'm proposing to change the level of this error log to error from debug


ylecuyer:
@googlebot I signed it!

emanuel-campos (approved):

bajajneha27:
chore(core): log errors as error instead of debug

Issue #690, opened 2018-06-15
Make the retry mechanism more configurable


Richard-Degenne:
Hi,

As you can see [here](https://github.com/google/google-api-ruby-client/blob/2a4efbf3b9c0e39255c574bfdf9ec57489857fcf/lib/google/apis/core/http_command.rb#L93-L96), it is currently possible to configure the number of retries through request options, but not other things such as the retry intervals, which are hard-coded, or [other `retriable` options](https://github.com/kamui/retriable#options).

Can something be done about that, or is it a policy to have those magic numbers as intervals?

kented:
While uploading large files to Youtube I was receiving 'Google::Apis::TransmissionError: execution expired'. After some digging I learned that retriable has a default max_elapsed_time of 15 minutes. I was able to adjust this configuration globally to solve my problem. 

Would be great to have some sort of documentation for this underlying mechanism and/or expose some level of configuration directly. The [context](https://github.com/kamui/retriable#contexts) feature of retriable might be useful for this.

snirbenyosef:
@kented 
where did you configure the max_elapsed_time ?
i׳m having the same problem

PR #8106, opened 2022-01-19
Support configuration of max elapsed time of requests

stanhu:
By default, the Retriable gem sets `max_elapsed_time` to 15
minutes. This can result in large transfers being aborted prematurely
rather than being retried. Make it possible to configure this value in
case the default is too low.

Closes https://github.com/googleapis/google-api-ruby-client/issues/690

### DIFF
--- a/google-apis-core/spec/google/apis/options_spec.rb
+++ b/google-apis-core/spec/google/apis/options_spec.rb
@@ -17,6 +17,7 @@ require 'google/apis/options'
 
 RSpec.describe Google::Apis::RequestOptions do
   let(:options) { Google::Apis::RequestOptions.new }
+  let(:defaults) { described_class.default }
 
   it 'should not merge nil values' do
     options.retries = 1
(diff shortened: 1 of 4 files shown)


bajajneha27:
Hi @stanhu ,
Thank you for creating this pull request. 
We generally do not accept PRs without the test cases. Could you please add tests as well? 
If you're not able to do that for some reason, we'll create a separate PR with the changes.

stanhu:
@bajajneha27 Added tests.

google-apis-core/spec/google/apis/options_spec.rb
```diff
@@ -37,4 +38,15 @@
   it 'should allow nil in merge' do
     expect(options.merge(nil)).to be_an_instance_of(Google::Apis::RequestOptions)
   end
+
+  it 'sets default values' do
+    expect(defaults.retries).to eq(5)
```
bajajneha27, line 43: This value should be [0](https://github.com/googleapis/google-api-ruby-client/blob/7f61f108cbd32b0c5eaac23768d42960e59998a0/google-apis-core/lib/google/apis/options.rb#L111)

bajajneha27:
> @bajajneha27 Added tests.

Thanks @stanhu for adding the tests.

google-apis-core/spec/google/apis/options_spec.rb
stanhu: I thought so too, but `spec_helper.rb` overrides this. 😄

bajajneha27 (approved):
LGTM

bajajneha27 (approved):
feat(core): Support for max elapsed time configuration.

PR #10026, opened 2022-06-09
feat(core): add few more errors as retriable errors

bajajneha27:


google-apis-core/lib/google/apis/errors.rb
```diff
@@ -66,7 +66,11 @@ class RedirectError < Error
     class ClientError < Error
     end
 
-    # A 4xx class HTTP error occurred.
+    # A 408 HTTP error occurred.
+    class RequestTimeOutError < ClientError
```
SandeepTuniki, line 70: Why does this inherit from the `ClientError` class while the others inherit from `Error` class?

bajajneha27: Since it's a `4xx` error, it makes sense to derive from [`ClientError`](https://github.com/googleapis/google-api-ruby-client/pull/10026/files#diff-9646a4866242f36eafecd780890cd92a622b53cf5013e7a9f343e0059796fe98R66)

NivedhaSenthil: it would also be backward compatible since earlier it was throwing ClientError

bajajneha27: Right. Thanks Nivedha

SandeepTuniki (approved):

PR #10542, opened 2022-06-28
chore: Add storage specific download code in order to respond with http response header

bajajneha27:


bajajneha27:
I haven't deliberately refactored the code and let it be redundant as we didn't want to spend too much time doing that.

google-apis-core/lib/google/apis/core/storage_download.rb
```diff
@@ -0,0 +1,88 @@
...
+        # @private
+        # @param [HTTPClient] client
+        #   HTTP client
+        # @yield [result, err] Result or error if block supplied
+        # @return [Object]
+        # @raise [Google::Apis::ServerError] An error occurred on the server and the request can be retried
+        # @raise [Google::Apis::ClientError] The request is invalid and should not be retried without modification
+        # @raise [Google::Apis::AuthorizationError] Authorization is required
+        def execute_once(client, &block)
+          request_header = header.dup
```
bajajneha27, line 38: I haven't deliberately refactored the code and let it be redundant as we didn't want to spend too much time doing that.

SandeepTuniki (approved):

NivedhaSenthil (approved):
I do not have much context otherwise implementation looks good to me

PR #11024, opened 2022-07-13
Remove rexml as a runtime dependency

ekohl:
This was added in 660a79807ab250b4e95179247e00ee8a364b9ce1 (see https://github.com/googleapis/google-api-ruby-client/issues/871 as well), but I can't reproduce the failure. I also can't find out where rexml was used in the past.

### DIFF
--- a/google-apis-core/google-apis-core.gemspec
+++ b/google-apis-core/google-apis-core.gemspec
@@ -26,6 +26,5 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "mini_mime", "~> 1.0"
   gem.add_runtime_dependency "googleauth", ">= 0.16.2", "< 2.a"
   gem.add_runtime_dependency "httpclient", ">= 2.8.1", "< 3.a"
-  gem.add_runtime_dependency "rexml"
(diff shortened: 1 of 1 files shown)


ekohl:
Before I sign any CLA I would like to know if this is even valid, or if I'm just missing something where it is needed.

bajajneha27:
Hi @ekohl 
Help me understand, are you saying that you tried removing this gem and you couldn't reproduce the failure? Can you please also tell me how you tried to check for failures?

ekohl:
I've tried to look at the git history to see why it was added. It clearly was added, likely because of some failure. However, now removing the dependency still passes the test suite so I'd think it's redundant. Having fewer dependencies is better so I'd like to see it dropped.

Does that clarify things?

bajajneha27:
Thanks for clarifying!
Can you please sign the CLA for us to proceed with the PR?

ekohl:
I would need to discuss that with legal. Is there a chance you could submit a similar PR with these changes instead?

bajajneha27:
Sure. We'll try to cover this in a separate PR. Closing this one. Thank you!

PR #11508, opened 2022-09-01
feat(core): Add storage upload to move away from unified upload protocol

bajajneha27:


google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,181 @@
...
+
+        # Ensure the content is readable and wrapped in an IO instance.
+        #
+        # @return [void]
+        # @raise [Google::Apis::ClientError] if upload source is invalid
+        def prepare!
+          @upload_url = nil
+          @offset = 0
+          @upload_incomplete = true
+          # Prevent the command from populating the body with form encoding, by
```
tritone, line 55: Why is this necessary?

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,181 @@
...
+        # @yield [result, err] Result or error if block supplied
+        # @return [Object]
+        # @raise [Google::Apis::ServerError] An error occurred on the server and the request can be retried
+        # @raise [Google::Apis::ClientError] The request is invalid and should not be retried without modification
+        # @raise [Google::Apis::AuthorizationError] Authorization is required
+        def execute(client)
+          prepare!
+          opencensus_begin_span
+
+          do_retry :initiate_resumable_upload, client
```
tritone, line 97: I assume the idempotency considerations for the retry are handled at the level of the manual storage client?

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,181 @@
...
+            do_retry :send_upload_command, client
+          end
+        ensure
+          opencensus_end_span
+          @http_res = nil
+          release!
+        end
+
+        def initiate_resumable_upload(client)
+          logger.debug { sprintf('Intiating resumable upload command to %s', url) }
```
tritone, line 108: Is there a standard for what gets logged in the Ruby client? Just curious since we don't really do debug logging in Go currently.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,181 @@
...
+        def send_upload_command(client)
+          logger.debug { sprintf('Sending upload command to %s', @upload_url) }
+
+          remaining_content_size = upload_io.size - @offset
+          current_chunk_size = remaining_content_size < CHUNK_SIZE ? remaining_content_size : CHUNK_SIZE
+
+          request_header = header.dup
+          request_header[CONTENT_RANGE_HEADER] = sprintf('bytes %d-%d/%d', @offset, @offset+current_chunk_size-1, upload_io.size)
+          request_header[CONTENT_LENGTH_HEADER] = current_chunk_size
+          body = upload_io.read(current_chunk_size)
```
tritone, line 144: What happens if a chunk upload fails and the input has already been read? Is the upload_io able to rewind? I don't really see how this happens in the code (but my ruby isn't very good!).

google-apis-generator/lib/google/apis/generator/templates/_method.tmpl
```diff
@@ -50,7 +50,15 @@
 # @raise [Google::Apis::ClientError] The request is invalid and should not be retried without modification
 # @raise [Google::Apis::AuthorizationError] Authorization is required
 def <%= api_method.generated_name %>(<% for param in api_method.required_parameters %><%= param.generated_name %>, <% end %><% if api_method.request %><%= api.schemas[api_method.request._ref].generated_name %>_object = nil, <% end %><% for param in api_method.optional_query_parameters %><%= param.generated_name %>: nil, <% end %><% for param in api.parameters.values.reject {|p| p.name == 'key'} %><%= param.generated_name %>: nil, <% end %><% if api_method.supports_media_upload? %>upload_source: nil, content_type: nil, <% elsif api_method.supports_media_download? %>download_dest: nil, <%  end  %>options: nil, &block)
-<% if api_method.supports_media_upload? -%>
+<% if api_method.supports_media_upload? && api.name == 'Storage' %>
+  if upload_source.nil?
+    command = make_simple_command(:<%= api_method.http_method.downcase %>, '<%= api_method.path %>', options)
+  else
+    command = make_storage_upload_command(:<%= api_method.http_method.downcase %>, '<%= api_method.path %>', options)
```
tritone, line 57: Does this need to be changed in the generated code in this PR also? In Go, presubmits will fail if the generated code doesn't match.

tritone (commented):
Generally looking good, nice progress here! A few things, mostly questions.

google-apis-generator/lib/google/apis/generator/templates/_method.tmpl
bajajneha27: Once this change is merged, the generator will generate the new code and raise a PR. Or I'll do it manually.

bajajneha27:
@tritone Integration tests are already there in [google-cloud-ruby](https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-storage/acceptance/storage/file_test.rb). I tested them against this code change and they ran successfully. The tests look thorough enough to me and covers all cases. So these tests along with conformance tests should be enough for our testing, I believe?

google-apis-core/spec/google/apis/core/storage_upload_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'spec_helper'
+require 'google/apis/core/storage_upload'
+require 'google/apis/core/json_representation'
+
+RSpec.describe Google::Apis::Core::StorageUploadCommand do
```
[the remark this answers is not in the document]
tritone: I think as long as the `google-cloud-ruby` integration tests pass with these changes, we are good.

google-apis-core/lib/google/apis/core/storage_upload.rb
tritone: Ah I wasn't sure how you decide which things to log using debug logging. Why here and not other places? Although maybe this is just copied from existing code?

google-apis-core/lib/google/apis/core/storage_upload.rb
tritone: Okay, gotcha -- so the rewinding happens in L153 I see

tritone (approved):
LGTM, pending approval from a Ruby reviewer and confirmation that the google-cloud-ruby storage integration tests pass.

google-apis-core/lib/google/apis/core/base_service.rb
```diff
@@ -318,6 +319,26 @@ def make_upload_command(method, path, options)
           command
         end
 
+        # Create a new storage upload command.
+        # This is specifically for storage because we are moving to a new upload protocol.
+        # Ref: https://cloud.google.com/storage/docs/performing-resumable-uploads
+        #
+        # @param [symbol] method
```
dazuma, line 326: `Symbol` is a class and should be capitalized

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+module Google
+  module Apis
+    module Core
+      # Base upload command. Not intended to be used directly
+      # @private
+      class StorageUploadCommand < ApiCommand
+        CONTENT_LENGTH_HEADER = 'Content-Length'
+        CONTENT_TYPE_HEADER = 'Content-Type'
+        UPLOAD_CONTENT_TYPE_HEADER = 'X-Upload-Content-Type'
+        LOCATION_HEADER = 'Location'
```
dazuma, line 30: I know we're not running rubocop on this repo yet, but let's start using our style for new code (e.g. double quotes for all strings, omit parens in method calls further down, etc.)

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+
+          super
+          if streamable?(upload_source)
+            self.upload_io = upload_source
+            @close_io_on_finish = false
+          elsif self.upload_source.is_a?(String)
+            self.upload_io = File.new(upload_source, 'r')
+            if self.upload_content_type.nil?
+              type = MiniMime.lookup_by_filename(upload_source)
+              self.upload_content_type = type && type.content_type
```
dazuma, line 69: I think `type&.content_type` is a bit more clear.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+          @offset += current_chunk_size if @upload_incomplete
+          success(result)
+        rescue => e
+          upload_io.pos = @offset
+          error(e, rethrow: true)
+        end
+
+        # Check the to see if the upload is complete or needs to be resumed.
+        #
+        # @param [Fixnum] status
```
dazuma, line 162: Fixnum is defunct as of Ruby 2.4. Use Integer.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+            @close_io_on_finish = false
+          elsif self.upload_source.is_a?(String)
+            self.upload_io = File.new(upload_source, 'r')
+            if self.upload_content_type.nil?
+              type = MiniMime.lookup_by_filename(upload_source)
+              self.upload_content_type = type&.content_type
+            end
+            @close_io_on_finish = true
+          else
+            fail Google::Apis::ClientError, 'Invalid upload source'
```
frankyn, line 73: It would help to know which type is not supported.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+          end
+          if self.upload_content_type.nil? || self.upload_content_type.empty?
+            self.upload_content_type = 'application/octet-stream'
```
frankyn, line 76: GCS handles the default value for users; `X-Goog-Upload-Content-Type` is optional unless that's changed recently.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,184 @@
...
+        # @return [HTTP::Message]
+        # @raise [Google::Apis::ServerError] Unable to send the request
+        def send_upload_command(client)
+          logger.debug { sprintf('Sending upload command to %s', @upload_url) }
+
+          remaining_content_size = upload_io.size - @offset
+          current_chunk_size = remaining_content_size < CHUNK_SIZE ? remaining_content_size : CHUNK_SIZE
+
+          request_header = header.dup
+          request_header[CONTENT_RANGE_HEADER] = sprintf('bytes %d-%d/%d', @offset, @offset+current_chunk_size-1, upload_io.size)
```
frankyn, line 145: IIUC, this code accepts a streamable type such as `STDIN` which may know the `upload_io.size`.

You can fall back to unknown sized resumable uploads: 
image.

source: https://cloud.google.com/storage/docs/streaming#rest-apis

google-apis-core/spec/google/apis/core/storage_upload_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+  include TestHelpers
+  include_context 'HTTP client'
+
+  let(:command) do
+    command = Google::Apis::Core::StorageUploadCommand.new(:post, 'https://www.googleapis.com/zoo/animals')
+    command.upload_source = StringIO.new('Hello world')
```
frankyn, line 25: Please include IO type which does not have a known size in your tests.

frankyn (changes requested):

google-apis-core/lib/google/apis/core/storage_upload.rb
bajajneha27: Discussed offline. We'll implement this in a separate PR. Created b/246693802 to track it.

google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -0,0 +1,181 @@
...
+          request_header[CONTENT_LENGTH_HEADER] = current_chunk_size
+          body = upload_io.read(current_chunk_size)
+
+          response = client.put(@upload_url, body: body, header: request_header, follow_redirect: true)
```
dazuma, line 146: Small nit here, but I'd name the `body` local variable something like `chunk_body` instead, to distinguish it from `self.body` which is a method on the base class (and is treated as such in line 120). This will protect you from a bug if you ever move these lines around and forget that the name is overloaded.

bajajneha27: Makes sense. Thanks for pointing it out.

frankyn (approved):

dazuma (approved):

PR #12454, opened 2022-10-21
fix(main): Updated if condition in release-generated

diptanshumittal:


.kokoro/release-generated.sh
```diff
@@ -11,7 +11,7 @@ gem install --no-document toys
 toys release install-python-tools -v
 # This is not called from autorelease, so don't run publish-reporter-script
 echo "${GEMS_LIST}"
```
shivgautam, line 13: Why do we need to print the gem list?

shivgautam (approved):

PR #13099, opened 2022-12-17
remove: Unused variable

johnmanjiro13:
The `length` variable is defined but never used.

This variable had been unused since 2017.
ref: https://github.com/googleapis/google-api-ruby-client/pull/564


dazuma (approved):
LGTM. Thanks!

PR #13213, opened 2023-01-12
fix: Improve upload performance

stanhu:
The change in #11508 introduced a significant performance
regression. It caused the HTTP client to read 8 MB chunks from the
upload source and send each chunk in separate PUT requests with a
String object instead of an IO object. This significantly slowed
upload performance since the SSL socket sends 16K at a time and
resizes the String buffer with each iteration. This caused CPU to
skyrocket since it requires allocating a new buffer and copying
existing data into that string. The garbage collector would have to
work hard to keep up.                                                                                                                                                            
                                                                                                                                                                                               
To eliminate this unnecessary String resizing, wrap the read chunk in                                                                                                                          
a `StringIO` object. This enables `OpenSSL::Buffering` to read and                                                                                                                             
write a full 16K buffer.                                                                                                                                                                       

Relates to #13212

### DIFF
--- a/google-apis-core/lib/google/apis/core/storage_upload.rb
+++ b/google-apis-core/lib/google/apis/core/storage_upload.rb
@@ -15,6 +15,7 @@
 require 'google/apis/core/http_command'
 require 'google/apis/core/api_command'
 require 'google/apis/errors'
+require 'stringio'
 require 'tempfile'
 require 'mini_mime'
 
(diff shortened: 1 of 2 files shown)


dazuma (changes requested):
Thanks for the analysis. This indeed looks like it should help. Can you also `require 'stringio'` at the top?

dazuma (approved):

PR #13405, opened 2023-01-31
fix: Updated README example

christiansaiki:
Using the latest Content API version as Google has sunset v2, hence you'll get a 404 error when using the old example. By using v2.1 you'll get the correct results.


bajajneha27 (approved):

bajajneha27:
Hi @christiansaiki 
Thanks for raising this PR. Can you please push a dummy commit so that the CI gets triggered?

dazuma (approved):

PR #17133, opened 2024-01-17
feat(generator): Update core dependency to at least 0.11.3

shivgautam:


shivgautam:
> Let's let #17132 update the dependency to 0.12.0.

Sounds good. I'll close this one. But that PR does not update the google-apis-generator.gemspec. Is that intended?

PR #19794, opened 2024-06-28
feat: Update API name definitions for Android Publisher API Subscriptionsv2

chrt00:
The Ruby client is missing the following REST endpoints.

- https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptionsv2/get 
- https://developers.google.com/android-publisher/api-ref/rest/v3/purchases.subscriptionsv2/revoke

I added the updated `api_names.yml` for the api generator.

I'm not a Google employee so I'm unable to check if this will pass the actions workflows.

The annotations may be weird as the suffix is `v2` and not delimited by a `.`
see Fix singularization of drives #809 for reference.


chrt00:
Found the definition - closing. https://github.com/googleapis/google-api-ruby-client/blob/b7484b91bb3f2e00224a0437a19d72868deed682/generated/google-apis-androidpublisher_v3/lib/google/apis/androidpublisher_v3/service.rb#L4037

PR #21648, opened 2025-02-22
Ensure compatibility with ruby --enable-frozen-string-literal

byroot:
Ref: https://bugs.ruby-lang.org/issues/20205

Since Ruby 2.4 it is possible to run ruby with `--enable-frozen-string-literal` and it's supposed to be the default behavior in the future.

It just require some very minor change for this gem to be compatible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,10 @@ jobs:
           - os: windows-latest
             ruby: "3.3"
             task: "--include-spec"
+          - os: ubuntu-latest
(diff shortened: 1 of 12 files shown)


wlipa:
It would be great if this could be released soon.  These warnings massively pile up in my logs under ruby 3.4.

byroot:
@dazuma sorry for the ping, but would you have time to look at this PR? This will become a compatibility issue in the near future.

dazuma (approved):
Thanks!

PR #22517, opened 2025-04-21
fear(core): setting idempotency header

shubhangi-google:
adding X-Goog-Gcs-Idempotency-Token


google-apis-core/lib/google/apis/core/api_command.rb
```diff
@@ -18,7 +18,7 @@
 require 'google/apis/errors'
 require 'json'
 require 'retriable'
-require "securerandom"
+require 'securerandom'
```
bajajneha27, line 21: Once we confirm that the `invocation_id` and the `idempotency-token` are same, we can remove this.

google-apis-core/lib/google/apis/core/api_command.rb
```diff
@@ -70,6 +70,7 @@ def initialize(method, url, body: nil, client_version: nil)
         def prepare!
           set_api_client_header
           set_user_project_header
+          set_idempotency_token_header if options.add_idempotency_token_header
```
bajajneha27, line 73: Can you please verify that the value is same when retry the same operation?

google-apis-core/lib/google/apis/options.rb
```diff
@@ -139,5 +141,6 @@ def merge(options)
     RequestOptions.default.quota_project = nil
     RequestOptions.default.add_invocation_id_header = false
     RequestOptions.default.upload_chunk_size = 100 * 1024 * 1024 # 100 MB
+    RequestOptions.default.add_idempotency_token_header = true
```
bajajneha27, line 144: Do we really need this option if we're enabling this by default?

google-apis-core/spec/google/apis/core/api_command_spec.rb
```diff
@@ -250,11 +263,17 @@
       command.options.add_invocation_id_header = true
       result = command.execute(client)
       invocation_id_header = command.header["X-Goog-Api-Client"]
-      
       expect(invocation_id_header).to include("gccl-invocation-id")
       expect(a_request(:get, 'https://www.googleapis.com/zoo/animals')
         .with { |req| req.headers['X-Goog-Api-Client'] == invocation_id_header }).to have_been_made.times(2)
     end
+
+    it 'should keep same idempotency_token across retries' do
```
bajajneha27, line 271: How are we retrying here?

google-apis-core/lib/google/apis/options.rb
shubhangi-google: I think we need this option
as the request being created in this test case https://github.com/googleapis/google-api-ruby-client/blob/cd1a7119dfb3ec3cb68a9c8ea4e0f45b50a04916/google-apis-core/spec/google/apis/core/service_spec.rb#L344-L350
is not adding the header by default if this option is not added
please help me with your understanding on this

bajajneha27: Question is, should it enabled by default or not.

shubhangi-google: it should be enabled by default

google-apis-core/spec/google/apis/core/api_command_spec.rb
```diff
@@ -250,11 +263,20 @@
...
+      command.options.add_idempotency_token_header = true
+      result = command.execute(client)
+      idempotency_token_header = command.header['X-Goog-Gcs-Idempotency-Token']
+      expect(command.header['X-Goog-Gcs-Idempotency-Token']).to eq(idempotency_token_header)
+      expect(a_request(:get, 'https://www.googleapis.com/zoo/animals')
```
bajajneha27, line 276: Looks like we're trying to different requests here. I'm a little confused.

shubhangi-google: we are checking retries on  same request

shubhangi-google: did not understand your concerns here

google-apis-core/lib/google/apis/options.rb
shubhangi-google: no we don't need this option I have updated the PR and removed it

google-apis-core/spec/google/apis/core/api_command_spec.rb
bajajneha27: The two commands are different, right?
So, to make sure we're calling the same operation multiple times. We should keep the command ( rather the code ) also same to avoid any confusion.

google-apis-core/lib/google/apis/options.rb
bajajneha27: I can still the option being used in multiple places. For ex.  [https://github.com/googleapis/google-api-ruby-client/pull/22517/files#diff-0b2335d085c2ef6552ca9cec9b610a712ca9a20b11c13da485c020014ea66359R72](https://github.com/googleapis/google-api-ruby-client/pull/22517/files#diff-0b2335d085c2ef6552ca9cec9b610a712ca9a20b11c13da485c020014ea66359R72)

bajajneha27: Oh! Why do we need this option ?

shubhangi-google: please correct me if my understanding is wrong

bajajneha27: Yes, that's what I had thought.

google-apis-core/spec/google/apis/core/api_command_spec.rb
shubhangi-google: please check if your concern here is resolved now
I have checked the retry test case execution same as we are doing for invocation_id header [29d078d6c679cfc2f6622d57cb22b3b0fa0e2b08](https://github.com/shubhangi-google/google-api-ruby-client/commit/29d078d6c679cfc2f6622d57cb22b3b0fa0e2b08)

bajajneha27 (approved):

PR #23018, opened 2025-05-06
Update README.md with Ruby 3.1+ support info.

simi:
- seen 3.1+ is still tested on CI
- 3.1 is EOL also, I'm happy to submit other PR to remove it from CI and gemspec

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ services.
 
 ## Supported Ruby versions
 
-This library is supported on Ruby 2.7+.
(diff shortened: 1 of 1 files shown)


dazuma (approved):
Thanks!

And yes, 3.1 is officially EOL but still supported in these libraries for a short while longer, and so it intentionally remains in the CI matrix. See https://cloud.google.com/ruby/getting-started/supported-ruby-versions for more info.

simi:
Do I need to update commit message?

PR #25157, opened 2025-12-12
feat(storage): adding checksum header

shubhangi-google:
This pull request introduces a mechanism to enable checksum validation during the finalization of object uploads. By including 'crc32c' and 'md5Hash' in the 'X-Goog-Hash' header for the last uploaded chunk, it ensures data integrity.

**Highlights:**
Checksum Validation: Implemented the addition of the 'X-Goog-Hash' header, containing 'crc32c' and 'md5Hash' values, specifically for the final chunk of an object upload to ensure data integrity.

### DIFF
--- a/google-apis-core/lib/google/apis/core/storage_upload.rb
+++ b/google-apis-core/lib/google/apis/core/storage_upload.rb
@@ -195,7 +195,7 @@ module Google
           success(result)
         rescue => e
           logger.warn {
-            "error occurred please use uploadId-#{response.headers['X-GUploader-UploadID']} to resume your upload , error==> #{e}"
+            "error occurred please use uploadId-#{response.headers['X-GUploader-UploadID']} to resume your upload"
           } unless response.nil?
           upload_io.pos = @offset
           error(e, rethrow: true)


google-apis-core/lib/google/apis/core/storage_upload.rb
```diff
@@ -191,7 +195,7 @@ def send_upload_command(client)
           success(result)
         rescue => e
           logger.warn {
-            "error occured please use uploadId-#{response.headers['X-GUploader-UploadID']} to resume your upload"
+            "error occurred please use uploadId-#{response.headers['X-GUploader-UploadID']} to resume your upload , error==> #{e}"
```
bajajneha27: Why are returning the error now?

shubhangi-google: was returning it to capture the errors in case of mismatched  header or token (while working on this functionality was facing issues with md5 encoding initially )
shall i undo it?
cc: @bajajneha27

bajajneha27: Let's keep it the way it was before for now. We'll add it if required.

shubhangi-google: sure , reverted the change

PR #26530, opened 2026-05-19
storage(feat): Add accept encoding gzip for media operation only

shubhangi-google:
This pull request adds the Accept-Encoding: gzip header to storage upload and download commands and includes corresponding unit tests.

Associated Storage library Pr: https://github.com/googleapis/google-cloud-ruby/pull/33896

### DIFF
--- a/google-apis-core/lib/google/apis/core/base_service.rb
+++ b/google-apis-core/lib/google/apis/core/base_service.rb
@@ -262,6 +262,7 @@ module Google
           end
           batch_command.execute(client)
         end
+        
         # Get the current HTTP connection
         # @return [Faraday::Connection]
         def client
--- a/google-apis-core/lib/google/apis/core/base_service.rb
+++ b/google-apis-core/lib/google/apis/core/base_service.rb
@@ -262,7 +262,7 @@ module Google
           end
           batch_command.execute(client)
         end
-        
+
         # Get the current HTTP connection
         # @return [Faraday::Connection]
         def client
--- a/google-apis-core/lib/google/apis/core/base_service.rb
+++ b/google-apis-core/lib/google/apis/core/base_service.rb
@@ -262,7 +262,6 @@ module Google
           end
           batch_command.execute(client)
         end
-
         # Get the current HTTP connection
         # @return [Faraday::Connection]
         def client
@@ -420,6 +419,7 @@ module Google
           verify_universe_domain!
           template = Addressable::Template.new(root_url + upload_path + path)
           command = StorageUploadCommand.new(method, template, client_version: client_version)
+          command.header["Accept-Encoding"] ||= "gzip"
           command.options = request_options.merge(options)
           apply_command_defaults(command)
           command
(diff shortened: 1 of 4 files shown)


google-apis-core/lib/google/apis/core/base_service.rb
```diff
@@ -421,6 +421,10 @@ def make_storage_upload_command(method, path, options)
           template = Addressable::Template.new(root_url + upload_path + path)
           command = StorageUploadCommand.new(method, template, client_version: client_version)
           command.options = request_options.merge(options)
+          command.options.header = command.options.header&.dup || {}
+          unless command.options.header.any? { |k, _| k.to_s.casecmp('accept-encoding') == 0 }
```
bajajneha27, line 425: Why are we doing this?

bajajneha27 (changes requested):

google-apis-core/lib/google/apis/core/base_service.rb
shubhangi-google: we are checking if the accept-encoding header already exists if not then we will add it
(casecmp is being used to check the header with respect to case senstivity)

bajajneha27: `header` is a hashmap, right? Can't we directly check `header['accept-encoding]` ?

shubhangi-google: Yes, header is a hash, but it's a standard Ruby Hash where keys are case-sensitive. Because HTTP header names are case-insensitive, a user might have already provided it as 'Accept-Encoding' or 'ACCEPT-ENCODING'.

If we used header['accept-encoding'] directly, it would return nil if the casing didn't match exactly. Using any? with casecmp ensures we detect the header regardless of how the user cased it, so we don't add a duplicate.

bajajneha27 (approved):

google-apis-core/spec/google/apis/core/service_spec.rb
```diff
@@ -192,6 +200,10 @@
       expect(command.query).to include('alt' => 'media')
     end
 
+    it 'should include Accept-Encoding header' do
+      expect(command.options.header['Accept-Encoding']).to eq('gzip')
```
aandreassa, line 204: Please add a test where you verify different capitalization (e.g. `ACCEPT-ENCODING`)

aandreassa (approved):
A comment, but LGTM otherwise.

google-apis-core/spec/google/apis/core/service_spec.rb
shubhangi-google: added

aandreassa (approved):

aandreassa:
@shubhangi-google plmk if you want me to merge this.

shubhangi-google:
> @shubhangi-google plmk if you want me to merge this.

yes please merge this PR

aandreassa:
@shubhangi-google we will need to keep an eye on the release then!

PR #26621, opened 2026-06-09
fix(google-apis-core): Relax retriable dependency constraint

### LINKED ISSUE #26620: Support newer versions of `retriable` gem
Here https://github.com/googleapis/google-api-ruby-client/blob/f488bb196b6a824d5b5a2d751f0b611d5143f23f/google-apis-core/google-apis-core.gemspec#L30
 change it to `>= 3.1`. No need to use any upper limitations (like `~>` or `<`), because someone will ask the same as I ask here one month later.

quartzmo:
Change retriable dependency constraint to '>= 3.1' to support 4.x major versions.

Manually verified the test suite runs successfully against retriable 4.1.1.

fixes: #26620

### DIFF
--- a/google-apis-core/google-apis-core.gemspec
+++ b/google-apis-core/google-apis-core.gemspec
@@ -27,5 +27,5 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'mini_mime', '~> 1.1'
   gem.add_runtime_dependency 'multi_json', '~> 1.11'
   gem.add_runtime_dependency 'representable', '~> 3.0'
-  gem.add_runtime_dependency 'retriable', '~> 3.1'
+  gem.add_runtime_dependency 'retriable', '>= 3.1'
 end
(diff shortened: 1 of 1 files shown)


aandreassa (approved):

google-apis-core/google-apis-core.gemspec
```diff
@@ -27,5 +27,5 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'mini_mime', '~> 1.1'
   gem.add_runtime_dependency 'multi_json', '~> 1.11'
   gem.add_runtime_dependency 'representable', '~> 3.0'
-  gem.add_runtime_dependency 'retriable', '~> 3.1'
+  gem.add_runtime_dependency 'retriable', '>= 3.1'
```
aandreassa: Maybe upper bound on <5?

quartzmo: Done, thanks!

aandreassa (approved):

vivekkj123:
Hi @aandreassa @quartzmo 

Can we support newer version of retriable (5.x) by adjusting the upper bound ?

Debian package of ruby-google-apis-core fails to build against latest retriable due to this

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1146301

PR #26624, opened 2026-06-10
chore(.toys): run bundle in external process to avoid Windows DLL collision

### LINKED ISSUE #26615: build: Windows CI failure on digest gem duplicate loading conflict
CI for `windows-latest` under Ruby 3.4 fails during `bundle install`:

```
RuntimeError: Digest::Base cannot be directly inherited in Ruby
     52: C:/hostedtoolcache/windows/Ruby/3.4.9/x64/bin/toys:36:in '<main>'
     ...
      6: C:/hostedtoolcache/windows/Ruby/3.4.9/x64/lib/ruby/gems/3.4.0/gems/bundler-4.0.13/lib/bundler/checksum.rb:17:in 'Bundler::Checksum.from_gem_package'
```

### Explanation

1. For Ruby 3.4.9, Bundler `4.0.13` resolves standard library dependencies from scratch and decides to fetch and compile the standard library gem `digest` (version `3.2.0`) from RubyGems.
2. During the checksum verification phase, Bundler requires `digest`. Because the gem path is active, it loads the gem's native C extension DLL (`digest.so` / `digest.dll`).
3. On Windows, loading duplicate DLLs of a preloaded C extension initializes a separate memory block, resulting in a different address for `Digest::Base`. The gem's extension fails the subclassing check when instantiating `Digest::SHA256.new`, throwing: `Digest::Base cannot be directly inherited in Ruby`.

### Proposed Fix

Upgrade the CI check from Ruby 3.4 to Ruby 4.0 per our current maintenance schedule. In Ruby 4.0, standard libraries like `digest` are fully extracted from the interpreter core, which eliminates the duplicate DLL collision on Windows since only the gem version is loaded.

Note that we must probably also add `ostruct` as a development dependency.

quartzmo:
fixes: #26615

### DIFF
--- a/.toys/ci.rb
+++ b/.toys/ci.rb
@@ -68,6 +68,10 @@ end
 
 def run_in_dir dir
   Dir.chdir dir do
+    puts
+    puts "Running bundle install in #{dir}...", :bold
+    exec ["bundle", "install"]
+
     if include_spec
       puts
       puts "Running spec in #{dir}...", :bold
--- a/.toys/ci.rb
+++ b/.toys/ci.rb
@@ -68,8 +68,7 @@ end
 
 def run_in_dir dir
   Dir.chdir dir do
-    puts
-    puts "Running bundle install in #{dir}...", :bold
+    puts "\nRunning bundle install in #{dir}...", :bold
     exec ["bundle", "install"]
 
     if include_spec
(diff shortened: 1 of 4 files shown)


.toys/ci.rb
```diff
@@ -68,6 +68,10 @@ def run
 
 def run_in_dir dir
   Dir.chdir dir do
+    puts
```
aandreassa: nit: remove first `puts`?

aandreassa (approved):

.toys/ci.rb
quartzmo: Fixed.

aandreassa (approved):

PR #27338, opened 2026-07-17
feat(storage): setting idempotency token header

shubhangi-google:
This pull request introduces support for sending an idempotency token header (X-Goog-Gcs-Idempotency-Token) in API commands, enabled by default via the new add_idempotency_token_header option.
linked storage library Pr : https://github.com/googleapis/google-cloud-ruby/pull/34956


cpriti-os (changes requested):
AI review: there are two issues.

1. **Global Default Scope Leak (Critical Issue):**
By setting `RequestOptions.default.add_idempotency_token_header = true` in `google/apis/options.rb`, this PR enables the GCS-specific `X-Goog-Gcs-Idempotency-Token` header globally for **all 400+ Google API clients** using `google-apis-core` (e.g., YouTube, Drive, Compute Engine, Calendar, Bigtable). This violates service abstraction boundaries and adds a 36-byte UUID overhead and confusing headers to every non-GCS request. 
**Fix:** Default `add_idempotency_token_header` to `false` in `google-apis-core` and explicitly enable it in the `google-apis-storage_v1` client generation or configuration.

2. **Redundant Memoization:**
In `set_idempotency_token_header`, the `@idempotency_token ||= SecureRandom.uuid` memoization is functionally dead code because of the early return on the line right before it:
```ruby
return if header.any? { |k, _| k.to_s.downcase == 'x-goog-gcs-idempotency-token' }
@idempotency_token ||= SecureRandom.uuid
```
Because the `header` Hash is mutated in place and retained across retries of the `ApiCommand` instance, the first network attempt injects the header into `header`. On subsequent retries, the loop evaluates `return if header.any? { ... }`, successfully finding the original key, and returning early. Therefore, `@idempotency_token` is never evaluated or used again across retries. 
**Fix:** You can simply do `header['X-Goog-Gcs-Idempotency-Token'] = SecureRandom.uuid`, as caching the UUID as an instance variable is unnecessary.

google-apis-core/lib/google/apis/core/api_command.rb
```diff
@@ -175,8 +176,15 @@ def set_user_project_header
...
 
+        def set_idempotency_token_header
+          return if options&.header&.any? { |k, _| k.to_s.downcase == 'x-goog-gcs-idempotency-token' }
+          return if header.any? { |k, _| k.to_s.downcase == 'x-goog-gcs-idempotency-token' }
+          header['X-Goog-Gcs-Idempotency-Token'] = SecureRandom.uuid
+        end
+
         def invocation_id_header
-          "gccl-invocation-id/#{SecureRandom.uuid}"
+          @invocation_id ||= SecureRandom.uuid
```
bajajneha27: Why are we making this an instance variable?

google-apis-core/lib/google/apis/core/api_command.rb
```diff
@@ -151,7 +152,7 @@ def set_api_client_header
             .map { |(a, b)| b }
             .join(' ')
             .split
-            .find_all { |s| s !~ %r{^gl-ruby/|^gdcl/} }
+            .find_all { |s| s !~ %r{^gl-ruby/|^gdcl/|^gccl-invocation-id/} }
```
bajajneha27: Why are we adding invocation_id header regex here?

google-apis-core/lib/google/apis/core/api_command.rb
```diff
@@ -175,8 +176,15 @@ def set_user_project_header
           header['X-Goog-User-Project'] = quota_project_id if quota_project_id
         end
 
+        def set_idempotency_token_header
+          return if options&.header&.any? { |k, _| k.to_s.downcase == 'x-goog-gcs-idempotency-token' }
+          return if header.any? { |k, _| k.to_s.downcase == 'x-goog-gcs-idempotency-token' }
```
bajajneha27: Can you please help me understand how `header` is populated?

google-apis-core/lib/google/apis/options.rb
```diff
@@ -140,5 +143,6 @@ def merge(options)
     RequestOptions.default.quota_project = nil
     RequestOptions.default.add_invocation_id_header = false
     RequestOptions.default.upload_chunk_size = 100 * 1024 * 1024 # 100 MB
+    RequestOptions.default.add_idempotency_token_header = false
```
bajajneha27: Is adding idempotency token going to be optional?

I thought it was going to be added by default.

google-apis-core/spec/google/apis/core/api_command_spec.rb
```diff
@@ -67,6 +66,14 @@
       expect(command.header['x-goog-api-client']).to be nil
     end
 
+    it 'should not accumulate duplicate gccl-invocation-id clauses when prepare! is called multiple times' do
+      command.options.add_invocation_id_header = true
+      command.prepare!
+      first_header = command.header['X-Goog-Api-Client']
```
bajajneha27: The test talks about checking duplicity of `gccl-invocation-id` while over here we're checking `X-Goog-Api-Client` .
Is this intentional?

google-apis-core/spec/google/apis/core/api_command_spec.rb
```diff
@@ -88,7 +95,41 @@
     it "should set the gccl-invocation-id to a random UUID" do
       command.options.add_invocation_id_header = true
       command.prepare!
-      expect(command.header["X-Goog-Api-Client"]).to include("gccl-invocation-id")
+      expect(command.header["X-Goog-Api-Client"]).to match(/gccl-invocation-id\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/)
```
bajajneha27: Same thing here. We're setting `X-Goog-Api-Client` to match with `gccl-invocation-id` value.

google-apis-core/spec/google/apis/core/api_command_spec.rb
```diff
@@ -88,7 +95,41 @@
...
+
+      expect(cmd1.header['X-Goog-Gcs-Idempotency-Token']).not_to eq(cmd2.header['X-Goog-Gcs-Idempotency-Token'])
+    end
+
+    it "should respect custom X-Goog-Gcs-Idempotency-Token in options.header regardless of casing" do
+      command.options.add_idempotency_token_header = true
+      command.options.header = { 'x-goog-gcs-idempotency-token' => 'my-custom-token' }
+      command.prepare!
+      expect(command.header['x-goog-gcs-idempotency-token']).to eql 'my-custom-token'
+      expect(command.header['X-Goog-Gcs-Idempotency-Token']).to be_nil
```
bajajneha27: Why would it be `nil` here?

bajajneha27 (changes requested):

google-apis-core/lib/google/apis/core/api_command.rb
shubhangi-google: We need to remove any existing gccl-invocation-id from the X-Goog-Api-Client header before appending it again. If prepare! is called multiple times (which can happen, for example, during retries), the invocation ID would just keep appending to the header over and over. Filtering it out here prevents duplicate entries.

google-apis-core/lib/google/apis/core/api_command.rb
shubhangi-google: without an instance variable, every retry would generate a brand new UUID. We use an instance variable to memoize the UUID.

google-apis-core/lib/google/apis/options.rb
shubhangi-google: as per this comment https://github.com/googleapis/google-api-ruby-client/pull/27338#pullrequestreview-5177891552
It has to be optional and default to false in core library,  It is being enabled by default inside the Storage API client specifically.

bajajneha27: My question is, do we need this option at all?
Isn't the requirement to always keep the header?

google-apis-core/lib/google/apis/core/api_command.rb
shubhangi-google: When a request is made for the first time, this hash doesn't have the idempotency token, so the code  populates it with a SecureRandom.uuid.
However, if the API call fails (e.g., due to a 503 Service Unavailable) and the library automatically retries the request, the prepare! method is called again. During this retry, the internal header hash still contains the UUID that was generated on the first attempt. Above line detects this and returns early, ensuring the exact same UUID is sent again on the retry.

google-apis-core/spec/google/apis/core/api_command_spec.rb
shubhangi-google: `expect(command.header['x-goog-gcs-idempotency-token']).to eql 'my-custom-token'
`
Confirms that the user's custom lowercase header ('x-goog-gcs-idempotency-token') was preserved with the custom value.
`expect(command.header['X-Goog-Gcs-Idempotency-Token']).to be_nil
` Confirms that the library didn't accidentally fall back to its default behavior and append a second, PascalCased header ('X-Goog-Gcs-Idempotency-Token'). If this were not nil, the final HTTP request would contain two conflicting idempotency headers, which could cause errors on the backend server.

google-apis-core/lib/google/apis/core/api_command.rb
bajajneha27: My question is how we're populating `header` hash ?

google-apis-core/lib/google/apis/core/api_command.rb
bajajneha27: As discussed offline, please do the required testing.

google-apis-core/spec/google/apis/core/api_command_spec.rb
shubhangi-google: `gccl-invocation-id ` is appended as a small clause inside the `X-Goog-Api-Client` header string
 the test checks that the final `X-Goog-Api-Client `string is perfectly identical to older header  and calling the prepare! twice have not added duplicate invocation-id

google-apis-core/spec/google/apis/core/api_command_spec.rb
shubhangi-google: here we are checking for the `invocation-id` substring to be present in `X-Goog-Api-Client`

PR #27476, opened 2026-07-27
feat(core): implement path validation and traversal checks in HttpCommand

quartzmo:
Implement dynamic runtime template parsing and validation inside HttpCommand to protect Discovery REST clients from path traversal and parameter injection exploits.

Specifically, this change:
- Rejects query (?) and fragment (#) characters in path parameter values.
- Rejects slashes (/) and dots (. or ..) in simple path variables (standard wildcards).
- Validates path traversals in reserved path variables (+ or #, double wildcards) using a segment-boundary traversal validation algorithm.

In addition to the validation logic, this commit includes:
- Ruby style modernization: cleaned up and auto-corrected string quote style to favor single quotes where appropriate.
- Formatted long stub requests to follow style guidelines.

Note: If the canonical specification is updated to favor the simpler alternative, this implementation can be easily simplified to "Fail always if dots are found in a double-wildcard value" by replacing the segment-traversal logic with a simple check for any "." or ".." substring.

### DIFF
--- a/google-apis-core/lib/google/apis/core/http_command.rb
+++ b/google-apis-core/lib/google/apis/core/http_command.rb
@@ -481,8 +481,20 @@ module Google
           end
         end
 
-        # Validate path parameters against template to prevent path traversal and injection exploits
-        # @raise [Google::Apis::Error] if validation fails
+        # Validates user-supplied path parameter values against the URL template specification
+        # to prevent directory traversal and parameter injection exploits.
+        #
+        # Validation Mechanism:
+        # 1. Identifies query/fragment injections: Rejects values containing '?' or '#' characters.
+        # 2. For simple variables (standard single-wildcard behavior):
+        #    - Rejects if the value contains '/' (cannot span multiple path segments).
+        #    - Rejects if the value is exactly '.' or '..'.
+        # 3. For reserved variables (reserved expansion like '+' or '#', double-wildcard behavior):
+        #    - Splits the value by slash ('/') using a `-1` limit to preserve empty trailing segments.
+        #    - Rejects if any segment is a directory traversal segment ('.' or '..').
+        #    - Rejects empty segments ('', meaning duplicate slashes '//' or trailing slashes).
+        #
+        # @raise [Google::Apis::Error] If any validation check fails.
         def validate_path_parameters!
           template_pattern = url.pattern
 
--- a/google-apis-core/lib/google/apis/core/http_command.rb
+++ b/google-apis-core/lib/google/apis/core/http_command.rb
@@ -361,9 +361,17 @@ module Google
         ].freeze
 
         # Pattern to scan for RFC 6570 URI template expressions enclosed in curly braces {...}.
-        # Captures:
-        # 1. An optional expansion operator prefix (one of: +, #, ., /, ;, ?, &).
-        # 2. A comma-separated list of variable definitions containing their names and optional modifiers.
+        #
+        # Regex Mechanics:
+        # - Matches literal '{' and '}' characters.
+        # - Capture Group 1: Captures a single prefix operator character if present (one of: +, #, ., /, ;, ?, &).
+        # - Capture Group 2: Captures all characters up to the closing brace ('[^}]+').
+        #
+        # Expected Capture Format:
+        # Capture Group 2 is expected to contain a raw string block representing one or more variable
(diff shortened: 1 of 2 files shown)


google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -83,15 +83,15 @@ def initialize(method, url, body: nil)
           self.url = url
           self.url = Addressable::Template.new(url) if url.is_a?(String)
           self.method = method
-          self.header = Hash.new
+          self.header =({})
```
torreypayne, line 87: I'm curious whether this was a Rubocop fix or was there a technical detail about instantiating a Hash one way vs the other?

google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -400,18 +415,16 @@ def opencensus_begin_span
...
           sent_size =
             if body.respond_to? :bytesize
               body.bytesize
             elsif body.nil?
               0
             end
-          if sent_size
-            @opencensus_span.put_message_event OpenCensus::Trace::SpanBuilder::SENT, 1, sent_size
-          end
+          @opencensus_span.put_message_event OpenCensus::Trace::SpanBuilder::SENT, 1, sent_size if sent_size
```
torreypayne, line 428: super nit: I actually like the original pattern (more explicit/readable) FWIW

torreypayne (approved):
This looks awesome! The only thing I'm wondering is if there are any test cases we could extract from the original vulnerability ask to ensure we have coverage. Otherwise LGTM!

google-apis-core/lib/google/apis/core/http_command.rb
quartzmo: Rubocop!

aandreassa: Can we fix the missing space? FWIW, I think`self.header = {}` is sufficient.

google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -484,6 +496,63 @@ def normalize_query_value(v)
...
+            end
+
+            if v[:reserved]
+              value_segments = value.split('/', -1)
+              value_segments.each do |seg|
+                if seg == '.' || seg == '..'
+                  raise Google::Apis::Error,
+                        "Path traversal segment #{seg.inspect} is not allowed in parameter #{var_name}: #{value}"
+                end
+                raise Google::Apis::Error, "Invalid path segment '' in parameter #{var_name}" if seg == ''
```
aandreassa, line 549: `Invalid path segment '' in parameter #{var_name}` doesn't include `#{value}` and the literal `''` can be a bit cryptic. 
WDYT about making it clearer:
```ruby
raise Google::Apis::Error, "Empty path segment (e.g. '//' or trailing '/') is not allowed in parameter #{var_name}: #{value}" if seg == ''

File: google-apis-core/spec/google/apis/core/http_command_spec.rb [hunk 4]
```diff
@@ -543,83 +547,231 @@ class SecretPayload
...
+        expect { command.execute(client) }.to raise_error(Google::Apis::Error, /contains invalid characters/)
+      end
+    end
+
+    context 'with reserved path variables' do
+      # Source template specification:
+      # Pattern: v1/{+parent}/indexes
+      # Variable constraints (RFC 6570 reserved expansion):
+      # - Allows slashes (multiple path segments)
+      # - Relative traversals within the segment boundary (e.g. 'a/b/../c') are allowed
```
efevans, line 709: NIT: looks like some remnants from the original design survived

google-apis-core/lib/google/apis/core/http_command.rb
```diff
@@ -484,6 +497,69 @@ def normalize_query_value(v)
...
+              value_segments = unescaped_value.split('/', -1)
+              value_segments.each do |seg|
+                if seg == '.' || seg == '..'
+                  raise Google::Apis::Error,
+                        "Value for #{var_name} must not contain segments that are exactly . or .."
+                end
+                raise Google::Apis::Error, "Invalid path segment '' in parameter #{var_name}" if seg == ''
+              end
+            else
+              if unescaped_value.include?('/')
```
efevans, line 552: Kinda an analog question to the .NET discovery lib work; this check isn't on the .NET side, do we need to skirt parity here?

PR #27557, opened 2026-08-06
feat: respect api_list_config exclusions in generate-updates

torreypayne:
Part 2 of stack to address b/542740217 (Apiary library generation errors preventing releases)

Source-Link: https://b.corp.google.com/issues/542740217

## Summary
Excludes `discoveryengine.v1`, `discoveryengine.v1alpha`, and `discoveryengine.v1beta` from client generation until the upstream Discovery doc issue is resolved.

## Changes
1. Added `discoveryengine.v1`, `discoveryengine.v1alpha`, `discoveryengine.v1beta` to `exclude:` list in `api_list_config.yaml`.
2. Updated `.toys/generate-updates.rb` to require `psych` and reject excluded APIs in `list_apis_versions` for both `--all` and explicit `requested` runs.

### DIFF
--- a/.toys/generate-updates.rb
+++ b/.toys/generate-updates.rb
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 require "uri"
+require "psych"
 
 desc "Run standard Google client generation."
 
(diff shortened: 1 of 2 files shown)


.toys/generate-updates.rb
```diff
@@ -75,15 +76,30 @@ def run
...
 def list_apis_versions
-  return requested.map { |request| request.split(":") } unless all
-  path = git_cache.find("https://github.com/googleapis/discovery-artifact-manager.git",
-                        path: "discoveries/index.json", update: true)
-  apis_versions = []
-  JSON.parse(File.read path)["items"].each do |item|
-    next unless item["preferred"] || gem_exists?(item)
-    apis_versions << [item["name"], item["version"]]
+  excluded = excluded_apis
+  if all
```
aandreassa, line 80: I dont think we need this (if/else) branching. If we want to regen them all, we should remove from the `exclude` list temporarily. WDYT?

torreypayne: I agree with you 100%; AG jumped ahead of me on pushing this draft up 🫠 

See https://github.com/googleapis/google-api-ruby-client/pull/27560/changes for updated changes (had to re-stack the PRs)

